PR #5, opened 2021-03-10
Revert changes that break belongsToMany relations

mjauvin:


bennothommo:
@mjauvin What's broken with the belongsToMany relations at this point?

mjauvin:
@bennothommo I had to different bugs just using the OFFLINE.Mall plugin (trying to save a category) and using the Blog plugin (trying to delete a category).

I think we should definitely apply this PR to revert previous changes until we stabilize this release. Once everything is stable and we get time again, we can revisit this with WAY MORE testing before pushing it live.

mjauvin:
And there is still Teranode's bug we couldn't reproduce yet (but it's out there)...

LukeTowers:
@mjauvin can you post the details on those bugs?

mjauvin:
For OFFLINE.Mall, whenever I save an existing Category, it just says "The URL has to be unique".

I don't really have more details, but the point is that it's causing breaks in new places over and over since I've introduced this change and I really think it is better to revert this for now.

mjauvin:
The OFFLINE.Mall bug can be tested here if you wish:

https://e94e5afc582f.octodock.com/backend
user: admin
pass: admin

PR #9, opened 2021-03-17
Take the extension from options into account when saving the image

mjauvin:
ref. https://github.com/octobercms/library/pull/546


mjauvin:
@LukeTowers This is required to allow converting images to other formats (e.g. webp). Otherwise the resulting file will have the correct extension but the wrong image format.

mjauvin:
In the mean time, the following code is required:
```
        Event::listen('system.resizer.processResize', function ($resizer, $tempPath) {
            // Get the resizing configuration
            $config = $resizer->getConfig();
            $options = array_get($config, 'options', []);

            list($base, $ext) = explode('.', $tempPath);
            $newPath = $base . '.' . array_get($options, 'extension', $ext);

            // Resize the image
            $resizedImageContents = \Winter\Storm\Database\Attach\Resizer::open($tempPath)
                ->resize($config['width'], $config['height'], $options)
                ->save($newPath);

            if ($newPath != $tempPath) {
                File::move($newPath, $tempPath);
            }
            
            // Prevent any other resizing replacer logic from running
            return true;
        });
```

mjauvin:
Any concerns for merging this @LukeTowers?

src/Database/Attach/Resizer.php
```diff
@@ -686,4 +686,9 @@ protected function getSizeByFit($maxWidth, $maxHeight)
 
         return [$optimalWidth, $optimalHeight];
     }
+
+    protected function getExtension($path)
```
bennothommo, line 690: @mjauvin could you add a simple docblock here for API completeness?

mjauvin:
Closed in favor of #74

PR #10, opened 2021-03-17
Allow for class aliases to be specified in plugin/module class loader

bennothommo:
This provides an `addAliases` method to the module/plugin class loader, allowing modules and plugins to provide aliases on initialization. This will assist with rebranding. Aliases are lazy-loaded, so that they won't be needed unless some functionality specifically requests them.

Usage:

```
use Winter\Storm\Support\ClassLoader;

app(ClassLoader::class)->addAliases([
    Winter\Test\Model\Test::class => OldLab\Test\Model\Test::class
]);
```

### DIFF
--- a/src/Support/ClassLoader.php
+++ b/src/Support/ClassLoader.php
@@ -240,7 +240,7 @@ class ClassLoader
      * Gets an alias for a class, if available.
      *
      * @param string $class
-     * @return void
+     * @return string|null
      */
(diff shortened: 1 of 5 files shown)


LukeTowers:
@bennothommo looks good, any way to add a unit test for this?

bennothommo:
@LukeTowers tests added :)

PR #11, opened 2021-03-17
Bound events to serializable closures

mjauvin:
ref. https://github.com/octobercms/library/pull/551


LukeTowers:
No longer necessary with 1.2.

PR #13, opened 2021-03-17
[WIP] Semver parser

jaxwilko:
PR to add a semvar parser to the library. This has been created for use with the plugin replace system to allow for specific version replacing, however it is a useful tool and maybe of use to developers generally so it would be best to include it within the Storm library.

The `Semvar` class allows for testing if a specific semvar version matches a composer style rule, i.e.

```php
Semvar::match('>=1.0 <1.4 || >=1.6', '1.6'); // true
Semvar::match('>=1.0 <1.4 || >=1.6', '1.5.0') // false
```

It supports `>`, `<`, `>=`, `<=`, `^`, `~` currently and I'm planning to add support for `*` and `-`, however I wanted to get some initial thoughts before I proceed.

Any comments greatly welcome!


src/Parse/Semvar.php
```diff
@@ -0,0 +1,229 @@
+<?php namespace Winter\Storm\Parse;
+
+/**
+ * Semvar parser to allow for version matching against Composer
+ * style rules
+ *
+ * @author Jack Wilkinson
+ */
+
```
LukeTowers:
```suggestion
```

src/Parse/Semvar.php
```diff
@@ -0,0 +1,229 @@
...
+class Semvar
+{
+    const LOGICAL_OR = '||';
+
+    const OPERATOR_GTE   = 'greaterThanEqual';
+    const OPERATOR_GT    = 'greaterThan';
+    const OPERATOR_LTE   = 'lessThanEqual';
+    const OPERATOR_LT    = 'lessThan';
+    const OPERATOR_TILDY = 'tildy';
```
LukeTowers: Should be `OPERATOR_TILDE = 'tilde';`

LukeTowers (changes requested):

src/Parse/Semvar.php
jaxwilko: That's my dyslexia showing 😂. I'll change it in the next push :)

jaxwilko:
I've now added support for range `-` and wild `*` operators. With this update the functionality of this new parser should be complete. Any suggestions before considering merging it?

Examples of new functionality:

```php
Semvar::match('1.5.*', '1.5.6'); // true
Semvar::match('1.5.* || 2.4.*', '1.6'); // false
Semvar::match('1.0 - 1.8', '1.6'); // true
Semvar::match('1.0 - 1.8 || 3.2 - 4.0', '2.3.0'); // false
```

bennothommo:
@jaxwilko Is there a particular reason that you didn't opt to use the [`composer/semver`](https://github.com/composer/semver) package? Unless I'm mistaken, the `Comparator` class they provide pretty much does what this PR is intending to do?

jaxwilko:
@bennothommo I did look into it, but the `Comparator` class doesn't support testing the complex rules with multiple constraints. However re-looking at the readme they also provid a `Semver` class which has some methods that may provide the functionality we need, I'll take a look in the morning and see if it'll give us what we need

bennothommo:
@jaxwilko No worries - I would imagine it would given that Composer uses this already for its constraint checks, so it should support anything that the `composer.json` files support.

composer.json
```diff
@@ -44,7 +44,8 @@
         "league/csv": "~9.1",
         "nesbot/carbon": "^2.0",
         "laravel/framework": "~6.0",
-        "laravel/tinker": "~2.0"
+        "laravel/tinker": "~2.0",
+        "composer/semver": "^3.2"
```
LukeTowers, line 48: This should probably just be done in the system module itself if it's not being used anywhere else in the application.

jaxwilko: https://github.com/wintercms/wn-system-module/pull/2

I've now closed this PR

PR #14, opened 2021-03-20
Change Auth Manager to not expose whether user exists or not on auth

dhawton:
Applying logic of https://github.com/wintercms/winter/issues/32 to authentication check as well. We should not expose whether a user exists or not to the login form. Can either apply the check here or do a try/catch in the backend module.


LukeTowers:
This has been brought up several times before, but this will not change here because it is already handled in AuthException: https://github.com/wintercms/storm/blob/develop/src/Auth/AuthException.php#L36-L40. The correct way to handle this is to disable debug mode in production.

Thanks for your interest in helping keep Winter secure though!

PR #15, opened 2021-03-23
Add support for namespace aliases in Class loader

bennothommo:
This will resolve the need to specify an aliases file in the plugin when doing plugin replacements. We can simply alias all classes of a certain namespace (ie. the replaced plugin's namespace) to be redirected to a new namespace (ie. the new plugin's namespace).

### DIFF
--- a/src/Support/ClassLoader.php
+++ b/src/Support/ClassLoader.php
@@ -169,6 +169,21 @@ class ClassLoader
     }
 
     /**
+     * De-register the given class loader on the auto-loader stack.
+     *
+     * @return void
+     */
+    public function unregister()
+    {
+        if (!$this->registered) {
+            return;
+        }
+
+        spl_autoload_unregister([$this, 'load']);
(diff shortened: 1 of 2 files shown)


LukeTowers:
@bennothommo what about this PR solves the problem we're facing with the plugin replacing where we specify individual classes? Is it just that the plugin replacer can use the name of the plugin being replaced to generate these aliases automatically?

bennothommo:
@LukeTowers sorry, I should've mentioned this is one of two parts - there's some changes I need to put forward for the main repo as well in the replaces branch.

You're correct - this provides a mechanism for replacing an entire namespace with another, so in the instance of plugins, it would take the plugin name of the replaced plugin, determine the namespace and then route all class requests within that namespace to the replacing plugin. It eliminates the need for defined aliases, and more to the point, doesn't require us to instantiate a separate file (`init.php`) which might contain other functionality.

PR #16, opened 2021-03-24
Update Symfony/Yaml to v4, add support for YAML processing

### LINKED ISSUE #13: Support newer versions of symfony/yaml
We should add a preprocessor to YAML files to prevent the breaking changes in newer versions of symfony/yaml from affecting the ecosystem (i.e. unquoted string keys that could be interpreted as numbers no longer work in newer versions).

Refs: octobercms/october#5271

bennothommo:
Fixes https://github.com/wintercms/winter/issues/13

This PR upgrades `symfony/yaml` to version 4.4 to match the rest of the Symfony components in Storm. To support old `version.yaml` file types, this also adds processor support for the YAML parser, and adds a pre-processor that will fix up most issues in old `version.yaml` files so that they correctly parse.

I've tried to cover most bases where the new Symfony YAML parser might trip up, but if anyone can think of other cases, please let me know so I can add them to the test case.


LukeTowers:
LGTM!

LukeTowers:
@bennothommo On second glance, shouldn't the VersionYamlProcessor be located in the system module? Makes it harder to test the processor logic here, but perhaps we can add a basic processor as a test fixture here and then the full version yaml processor test in the winter repo instead. Also the VersionManager needs to set the processor in order for this to take effect anyways.

bennothommo:
@LukeTowers sounds good - will sort that out, although I won't put a test in for Storm as we can assume Yaml is adequately tested by Symfony, and I can't think of another processor we need right now.

PR #17, opened 2021-03-24
[WIP] Translator afterResolve event

jaxwilko:
Adds a new event that gets fired after an a translation has been resolved. Acts identically to beforeResolve except for it also passes the found line into the event, allowing for modifications to be made and fallback handlers for missing lines to be created.

Example:
```php
Event::listen(
    'translator.afterResolve',
    function (string $key, array $replace, ?string $line, ?string $locale) {
        // handle things 
        return $someVar;
    }
);
```

### DIFF
--- a/src/Translation/Translator.php
+++ b/src/Translation/Translator.php
@@ -127,7 +127,7 @@ class Translator extends TranslatorBase
             return $afterResolve;
         }
 
-        // If the line doesn't exist, we overrideswill return back the key which was requested as
+        // If the line doesn't exist, we will return back the key which was requested as
         // that will be quick to spot in the UI if language keys are wrong or missing
         // from the application's language files. Otherwise we can return the line.
         if (!isset($line)) {
--- a/src/Translation/Translator.php
+++ b/src/Translation/Translator.php
@@ -59,10 +59,8 @@ class Translator extends TranslatorBase
          *     });
          *
          */
-        if (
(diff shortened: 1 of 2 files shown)


src/Translation/Translator.php
```diff
@@ -105,7 +107,27 @@ public function get($key, array $replace = [], $locale = null, $fallback = true)
...
+         *
+         */
+        if (
+            isset($this->events)
+            && ($afterResolve = $this->events->fire('translator.afterResolve', [$key, $replace, $line, $locale], true))
+        ) {
+            return $afterResolve;
+        }
+
+        // If the line doesn't exist, we overrideswill return back the key which was requested as
```
mjauvin:
```suggestion
        // If the line doesn't exist, we will return back the key which was requested as
```

jaxwilko: Good spot, fix has been pushed

src/Translation/Translator.php
```diff
@@ -59,8 +59,10 @@ public function get($key, array $replace = [], $locale = null, $fallback = true)
          *     });
          *
          */
-        if (isset($this->events) &&
-            ($line = $this->events->fire('translator.beforeResolve', [$key, $replace, $locale], true))) {
```
LukeTowers: This is our current style for handling multi-line conditionals, please go back to it

PR #19, opened 2021-03-26
[WIP] Config Aliases

jaxwilko:
Adds support for config aliasing.

```php
Config::package('winter.test', '/path/to/config');
Config::registerNamespaceAlias('winter.test', 'winter.alias');

Config::get('winter.test::foo'); // bar
Config::get('winter.alias::foo'); // bar
```

### DIFF
--- a/src/Config/FileLoader.php
+++ b/src/Config/FileLoader.php
@@ -27,6 +27,13 @@ class FileLoader implements LoaderInterface
     protected $hints = [];
 
     /**
+     * All of the namespace aliases.
+     *
+     * @var array
+     */
+    protected $aliases = [];
+
+    /**
      * A cache of whether namespaces and groups exists.
      *
      * @var array
@@ -159,6 +166,13 @@ class FileLoader implements LoaderInterface
         if ($this->files->exists($path)) {
             $items = array_merge($items, $this->getRequire($path));
         }
+        // If package is has an alias and the package does not have a global config
+        elseif (isset($this->aliases[$package])) {
+            $path = $this->getPackagePath($this->aliases[$package], $group);
+            if ($this->files->exists($path)) {
+                $items = array_merge($items, $this->getRequire($path));
+            }
+        }
 
         // Once we have merged the regular package configuration we need to look for
         // an environment specific configuration file. If one exists, we will get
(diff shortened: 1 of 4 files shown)


LukeTowers:
Does this work with loading config from the alias location? I.e. if config/rainlab/blog/config.php exists and config/winter/blog/config.php doesn't, we want to make sure both rainlab.blog::item and winter.blog::item loads from config/rainlab/blog/config

jaxwilko:
We would handle that during plugin registration. However specifically with replaced plugins (at least with lang) we assume that the plugin is a fork and we don't provide fallback support to the original plugins files.

If there is a good reason for the replacing plugin to fallback to the original plugins config (if exists) if the replacing plugin does not include a config then we can handle this, but we'd do it during plugin registration (same place as we handle lang namespace aliasing).

I guess the question is do we want to handle it like the translator or do we want provide fallback support?

Either way, the code in storm should stay the same but we'll change how we use the `registerNamespaceAlias` in winter.

LukeTowers:
While it would technically be better to have both have full fallback support for the original plugin's overrides, the consequences of not having that support for overridden translations are much less than the consequences of not having that support for overridden config values; which could theoretically compromise the stability or security of the project if they don't update those config files properly when switching plugins.

All that to say, if it can be done in the system module instead then I'm fine with that, although I don't think we need to put any work into supporting the same logic for the translator; unless it's super easy to do and doesn't have any negative effects.

mjauvin:
@LukeTowers, @jaxwilko was referring to the plugin's ***default*** lang/config files... not the ***overrides*** under the app's lang or config folders. He's working at adding this, we all agree this is required if we want to seamlessly support migration to winter's rebranded plugins.

LukeTowers:
Gotcha. There's definitely no expectation from me to look at plugins/rainlab/blog/config when rainlab.blog is aliased to winter/blog.

jaxwilko:
To document the final additions of this PR, 2 new methods have been added to `Repository` which is access as `Config` inside Winter.

> registerNamespaceAlias

```php
Config::registerNamespaceAlias('winter.demo', 'alias.demo');
```
Will redirect all calls to alias.demo to winter.demo.

I.e.

```php
<?php
// plugins/winter/demo/config/config.php
return [
    'foo' => 'bar'
];
```

`config('alias.demo::foo');` returns `bar`

> registerPackageFallback

```php
Config::registerPackageFallback('winter.demo', 'alias.demo');
```
This will add the global override for `alias.demo` to the `winter.demo` namespace if there is a global config for alias.demo and not for winter.demo.

PR #21, opened 2021-03-28
[WIP v1.2] Add full compatibility with Symfony\Yaml 4.4

mjauvin:


mjauvin:
@LukeTowers can you merge this to fix the yaml files parsing?

Thanks.

LukeTowers:
Can you add tests for what happens if single and double quotes appear in the unquoted values?

mjauvin:
@LukeTowers You mean if a value in the original file (before it gets pre-parsed) contains single or double quotes?

tests/fixtures/yaml/symfony3.yaml
```diff
@@ -29,3 +29,13 @@ form:
             options:
                 true: true
                 false: false
+        testClass:
+            My\Class\Path: value
+        testQuoting:
+            complex1: 'this is "very" complex quoting'
+            complex2: "this is also 'very' complex quoting"
+            middle: this string is "quoted" in the middle
```
LukeTowers, line 37: How about the above but also just a single quote? Complex2 would pass without being modified because it's a quoted value in it's entirety, and middle would pass because the value would be wrapped in single quotes, but I'm wondering about `middeSingle: this string has some single quote's, and maybe a 'air quote', anything to make your life diff'cult`

LukeTowers:
@mjauvin what's the status on this?

mjauvin:
> @mjauvin what's the status on this?

@LukeTowers since we reverted @bennothommo 's work on this, it's currently irrelevant as far as I remember.

LukeTowers:
@bennothommo what's the status of this?

bennothommo:
@LukeTowers you'd have to check with @mjauvin of what the purpose of this PR was? I'm not sure if it's needed, but I also can't recall the reason we had to roll back my processor originally - there was some sort of issue with it, I imagine?

mjauvin:
@LukeTowers it was needed at the time (see the tests that were added). Not sure if it's still relevant.

PR #25, opened 2021-04-22
Resolve other issues with replaced plugin

mjauvin:
Moving the class_alias if block first resolves the loader issue when replaced plugin is still installed but disabled

### DIFF
--- a/src/Database/Relations/AttachOneOrMany.php
+++ b/src/Database/Relations/AttachOneOrMany.php
@@ -39,7 +39,14 @@ trait AttachOneOrMany
     public function addConstraints()
     {
         if (static::$constraints) {
-            $this->query->where($this->morphType, $this->morphClass);
+            if (starts_with($this->morphClass, 'Winter\\')) {
+                $this->query->where(function ($q) {
+                    $q->where($this->morphType, $this->morphClass);
+                    $q->orWhere($this->morphType, str_replace('Winter\\', 'RainLab\\', $this->morphClass));
+                });
+            } else {
+                $this->query->where($this->morphType, $this->morphClass);
+            }
 
             $this->query->where($this->foreignKey, '=', $this->getParentKey());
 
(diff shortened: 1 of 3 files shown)


src/Support/ClassLoader.php
```diff
@@ -98,6 +98,10 @@ public function __construct(Filesystem $files, $basePath, $manifestPath)
      */
     public function load($class)
     {
+        if (!is_null($alias = $this->getAlias($class))) {
```
LukeTowers, line 103:
```suggestion
        // Allow class aliases defined with addAliases() and addNamespaceAliases() to take
        // precedence over classes that actually exist on the filesystem
        if (!is_null($alias = $this->getAlias($class))) {
```

mjauvin:
@LukeTowers at first glance, do you see a problem running this check first?

LukeTowers:
No, just so long as we make it clear with those two methods that they have the power to prevent a class that actually exists but hasn't been loaded yet from being loaded; since class_alias() is typically a fallback method instead of an overriding one

src/Database/Relations/AttachOneOrMany.php
```diff
@@ -39,7 +39,14 @@ public function isPublic()
...
         if (static::$constraints) {
-            $this->query->where($this->morphType, $this->morphClass);
+            if (starts_with($this->morphClass, 'Winter\\')) {
+                $this->query->where(function ($q) {
+                    $q->where($this->morphType, $this->morphClass);
+                    $q->orWhere($this->morphType, str_replace('Winter\\', 'RainLab\\', $this->morphClass));
+                });
+            } else {
+                $this->query->where($this->morphType, $this->morphClass);
+            }
```
LukeTowers: I'm not comfortable with having this in the core, it's pretty hacky to have a vendor-specific conditional in the core and there could be a number of unforeseen issues with invisibly grabbing more data than expected by the developer.

mjauvin: isn't that what you meant by:

NOTE: We could also explore the idea of having Laravel replace these values at the query level seeing as it's highly likely that other plugins could be extending the replaced plugins and referencing those models in their own data structure.

LukeTowers: I more so meant doing so at the plugin level through an API of some sort, but wasn't quite sure how it'd work or if it'd even be feasible.

PR #26, opened 2021-04-25
upgrade to laravel 8

tschallacka:
Based on issue:
https://github.com/wintercms/winter/issues/23

All conflicts are resolved, all tests seem to pass without issue.
Some care is needed to review the wintercms implementation for the TransportAdapter as this has changed in Laravel 7 to MailAdapter. The tests pass, but integration may not, but I'm not familiar enough with the mail adapter to make an integration test.


tests/.phpunit.result.cache
```diff
@@ -0,0 +1 @@
...t::testThemesPath";d:0;s:25:"HelpersTest::testTempPath";d:0;s:28:"HelpersTest::testUploadsPath";d:0;s:26:"HelpersTest::testMediaPath";d:0;s:22:"MailFakeTest::testSend";d:0;s:23:"MailFakeTest::testQueue";d:0;s:35:"MailFakeTest::testIndexedArrayViews";d:0;s:33:"MailFakeTest::testNamedArrayViews";d:0;s:42:"MailFakeTest::testIndexedArrayViews_Queued";d:0;s:40:"MailFakeTest::testNamedArrayViews_Queued";d:0;s:31:"UrlGeneratorTest::testSimpleUrl";d:0;s:32:"UrlGeneratorTest::testComplexUrl";d:0;s:34:"UrlGeneratorTest::testReplacements";d:0;s:34:"UrlGeneratorTest::testJoinSegments";d:0;s:35:"UrlGeneratorTest::testStripSegments";d:0;s:39:"TranslatorTest::testSimilarWordsParsing";d:0;s:26:"TranslatorTest::testChoice";d:0.001;s:35:"TranslatorTest::testChoiceSublocale";d:0;s:50:"TranslatorTest::testOverrideWithBeforeResolveEvent";d:0.001;s:37:"TranslatorTest::testNamespaceAliasing";d:0;s:46:"TranslatorTest::testMixedCaseNamespaceAliasing";d:0;s:38:"EmailValidationTest::testDefaultFilter";d:0.001;s:34:"EmailValidationTest::testRFCFilter";d:0.001;s:30:"RuleObjectTest::testRuleObject";d:0;s:36:"RuleObjectTest::testRuleObjectPasses";d:0;s:47:"RuleObjectTest::testRuleObjectTranslatedMessage";d:0;}}}
```
LukeTowers: This should probably be added to gitignore

tests/Database/QueryBuilderTest.php
```diff
@@ -210,9 +210,7 @@ protected function getConnection($connection = null)
                 'rollBack',
                 'transactionLevel',
                 'pretend',
-            ])
-            ->addMethods([
-                'getDatabaseName',
+                'getDatabaseName'
```
LukeTowers, line 213: Could you explain this change?

src/Database/Query/Grammars/PostgresGrammar.php
```diff
@@ -1,6 +1,6 @@
 <?php namespace Winter\Storm\Database\Query\Grammars;
 
-use Winter\Storm\Database\QueryBuilder;
+use Illuminate\Database\Query\Builder;
```
LukeTowers: Could you explain this change? Also, upsert support was added to 8.x, so we should be able to test and then remove it from here as this was a backport.

src/Database/Pivot.php
```diff
@@ -68,10 +68,10 @@ public function __construct(ModelBase $parent, $attributes, $table, $exists = fa
     /**
      * Set the keys for a save update query.
      *
-     * @param  \Illuminate\Database\Eloquent\Builder
+     * @param  \Illuminate\Database\Eloquent\Builder $query
      * @return \Illuminate\Database\Eloquent\Builder
      */
-    protected function setKeysForSaveQuery(BuilderBase $query)
+    protected function setKeysForSaveQuery($query)
```
LukeTowers, line 74: Could you explain this change?

composer.json
```diff
@@ -39,11 +39,11 @@
         "linkorb/jsmin-php": "~1.0",
         "wikimedia/less.php": "~3.0",
         "scssphp/scssphp": "~1.0",
-        "symfony/yaml": "^3.4",
+        "symfony/yaml": "^5.1",
         "twig/twig": "~2.0",
         "league/csv": "~9.1",
         "nesbot/carbon": "^2.0",
-        "laravel/framework": "~6.0",
+        "laravel/framework": "^8.0",
```
LukeTowers: Why the change from ^ to ~?

composer.json
```diff
@@ -39,11 +39,11 @@
         "linkorb/jsmin-php": "~1.0",
         "wikimedia/less.php": "~3.0",
         "scssphp/scssphp": "~1.0",
-        "symfony/yaml": "^3.4",
+        "symfony/yaml": "^5.1",
```
LukeTowers, line 42: What changes are there in YAML 5 vs 4? I know there were breaking changes in 3-4 that we're addressing with YAML processing, but is there anything we should be concerned about in 4-5?

tschallacka:
Concerning https://github.com/wintercms/storm/pull/21   laravel 8 needs yaml 5.1 so the other branch referenced here should aim to optimize for 5.1.

LukeTowers (changes requested):

tests/Database/QueryBuilderTest.php
tschallacka: getDatabaseName is part of the interface. Phpunit complained it can't create a method that's already defined. 
By moving it to the array of methods it could be instantiated.

src/Database/Query/Grammars/PostgresGrammar.php
tschallacka: yes, you cannot change the contract on a method. Overriding the typehint of the parent class throws a fatal error.  See the related commit message https://github.com/wintercms/storm/pull/26/commits/ad9f2eaa26ae4f4cf856d86c4764f2da4b9d42fb

Also I was wondering if it could be removed as the methods are identical. I wasn't 100% sure what the reason was for them existingso I opted for keeping them. When in doubt don't touch it heh.

composer.json
tschallacka: For whatever reason my composer refused to perform updates to 8.0 without the ^. -vvv gave no meaningful insights. 
I tried this and it worked, I can change it to ~ if you want.

The [composer documentation](https://getcomposer.org/doc/articles/versions.md) also mentions

> This is the recommended operator for maximum interoperability when writing library code.

So I kind of use this default in my own projects. Please let me know what your preferred method is, they are basically the same.

composer.json
tschallacka: I have not looked into changes, I have no overview over breaking changes. Are there unit tests that adress stuff that break? It passed all current unit tests in storm? I assumed the current unit tests address all current stress points in the library? If you have an overview of bug reports of yaml that breaks I can take a targetted look.

Do the breaking changes only occur in combination with wintercms? Or certain plugins within wintercms?

src/Database/Pivot.php
tschallacka: The parent class doesn't have a query builder in the method signature.  Php threw a fatal error. See https://softwareengineering.stackexchange.com/questions/274118/php-extended-class-method-requires-same-signature-including-object-class-require/274262#274262

tests/.phpunit.result.cache
tschallacka: I would think you're right. I'll fix tgat tomorrow

src/Database/Query/Grammars/PostgresGrammar.php
LukeTowers: @tschallacka 

> Also I was wondering if it could be removed as the methods are identical. I wasn't 100% sure what the reason was for them existing so I opted for keeping them. When in doubt don't touch it heh.

100% 😄

So long as the methods are identical it should be fine to remove. The upsert methods were added because it was only added to 8.x but I needed it available in Winter using 6.x.

composer.json
LukeTowers: The breaking changes from 3 to 4 means that the usual version.yaml in use fails:

```yaml
1.1.2: !!! Important update
```
needs to be 
```yaml
"1.1.2": "!!! Important update"
```
in YAML 4. I haven't looked at any changes from 4 to 5 though.

src/Database/Query/Grammars/PostgresGrammar.php
tschallacka: I'll go ahead and remove it then.

I would suggest that backports in the future be flagged as such by a comment, so when the framework updates it will be easier to find them and to clean them up.

tschallacka:
The PHP 7.2 tests will fail as Laravel 8 is PHP 7.3 minimum requirement and Laravel 9 will be PHP 7.4. Do you want to enforce the PHP 7.4 or keep it at 7.3 and move to 7.4 when Laravel 9 is there?

LukeTowers:
@tschallacka you should be able to edit the github action workflows in the PR too. We should be targeting Laravel 9 with this branch so just target 7.4 as the minimum

mjauvin:
Yeah, probably better to update to 7.4 when we actually get to Laravel 9

composer.json
tschallacka: Hmmh what i've found is that when a string values starts with an exclamation mark `https://github.com/symfony/yaml/blob/afe64a9ce7718524116c9a3b0913e5163b49c405/Inline.php#L585` it will check if it matches a scalar, and that will break it. 

Valid scalars are 

```php
$valid_scalar_tags = [
                      '!!str ',
                      '! ',
                      '!php/object',
                      '!php/const',
                      '!!float ',
                      '!!binary '
];
```
So the string should be wrapped with quotes when it doesn't match a valid scalar, something in the line of

```php
public function checkValue($text)
{
  if(strpos($text, '!') === 0) {
    $valid_scalar_tags = [
                          '!!str ',
                          '! ',
                          '!php/object',
                          '!php/const',
                          '!!float ',
                          '!!binary '
    ];
    $lower = strtolower($text);
    foreach($valid_scalar_tags as $scalar) {
      if(strpos(lower, $scalar) === 0) {
        return $text;
      }
    }   
  return '"'.$text.'"';
  }
return $text;
}
```

or put a `! ` in front of it so it just gets substringed out and not considered as a scalar, so you don't have to worry about quotes https://github.com/symfony/yaml/blob/afe64a9ce7718524116c9a3b0913e5163b49c405/Inline.php#L589

I'm not sure which class handles the version.yalm loading, and I honestly don't have the time to write a pre processor or specialised class for this use case where a scalar interpretation is attempted.

I added a version.yaml test in the unit tests, and that works properly, the keys don't need to be quoted in 5.1 to be interpreted correctly as far as I can see, so it's only the scalar tags that cause issue in your provided example.

.editorconfig
```diff
@@ -1,7 +0,0 @@
-[*]
-end_of_line = lf
-charset = utf-8
-trim_trailing_whitespace = true
-insert_final_newline = true
-indent_style = space
-indent_size = 4
```
LukeTowers: I don't think you meant to delete this

composer.json
LukeTowers: We wouldn't want the version.yaml pre-processing to be done in the core anyways, see https://github.com/wintercms/winter/commit/58664605ea53465422882a1b543bc9d7a32b755c for how it's handled in the system module instead.

tschallacka: Do you want me to remove the version.yaml unit test then from this library?

.editorconfig
tschallacka: You're right, I had to reset my eclipse project, deleted that in my overzealousness in resetting the project base. Fixed it :-)

tschallacka:
@LukeTowers are you able to run phpunit tests on this library(1.1, 1.2/wip) without the overloading fatal errors I fixed in this pull request? Or are my computers just misconfigured?

LukeTowers:
@tschallacka are you talking about the testing suite on all the branches without this PR included? In that case, yes, all of them seem to be passing at the moment.

.editorconfig
```diff
@@ -4,4 +4,4 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 indent_style = space
-indent_size = 4
+indent_size = 4
```
LukeTowers:
```suggestion
indent_size = 4

```

.github/workflows/tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       max-parallel: 6
       matrix:
         operatingSystem: [ubuntu-latest, windows-latest]
-        phpVersion: ['7.2', '7.3', '7.4', '8.0']
+        phpVersion: ['7.3', '7.4', '8.0']
```
LukeTowers: might as well just get rid of 7.3 support altogether if Laravel 9 isn't going to support it

composer.json
LukeTowers: can we target dev-master or is that going to be too much of a headache?

src/Database/Pivot.php
LukeTowers: Gotta love poorly explained changes, modified in 8.7.0: https://github.com/laravel/framework/commit/77db028225ccd6ec6bc3359f69482f2e4cc95faf

src/Database/Traits/Encryptable.php
```diff
@@ -119,6 +117,6 @@ public function getEncrypter()
      */
     public function setEncrypter(\Illuminate\Contracts\Encryption\Encrypter $encrypter)
     {
-        $this->encrypter = $encrypter;
+        parent::encryptUsing($encrypter);
```
LukeTowers: Got a reference for this change?

src/Events/Dispatcher.php
```diff
@@ -30,7 +30,7 @@ class Dispatcher extends BaseDispatcher
      * @param  int  $priority
      * @return void
      */
-    public function listen($events, $listener, $priority = 0)
+    public function listen($events, $listener = null, $priority = 0)
```
LukeTowers, line 38: We'll need to implement / make sure our dispatcher supports the changes made in https://github.com/laravel/framework/commit/46084d946cdcd1ae1f32fc87a4f1cc9e3a5bccf6

src/Foundation/Exception/Handler.php
```diff
@@ -38,10 +38,10 @@ class Handler extends ExceptionHandler
      *
      * This is a great spot to send exceptions to Sentry, Bugsnag, etc.
      *
-     * @param  \Exception  $exception
+     * @param  \Throwable  $exception
```
LukeTowers: We should be able to get rid of the exception import at the top of the file, and we'll need to update the remaining references to \Exception (especially in the event docblocks) in this class.

composer.json
LukeTowers: That's probably fine to do yeah

tests/Database/Traits/EncryptableTest.php
```diff
@@ -22,7 +22,7 @@ public function setUp(): void
     public function testEncryptableTrait()
     {
         $testModel = new TestModelEncryptable();
-        $testModel->setEncrypter($this->encrypter);
+        $testModel->encryptUsing($this->encrypter);
```
LukeTowers: This should probably remain using the existing interface since that's the public API for the encryptable trait

tschallacka:
> @tschallacka are you talking about the testing suite on all the branches without this PR included? In that case, yes, all of them seem to be passing at the moment.

Yes. When I run `composer test` with any branch I get this output, so I'm curious about no other errors about incompatible child implementations of contracts.

console output

LukeTowers:
@tschallacka it's probably because you need to run composer update to swap your included version of Laravel from 8 back to 6 which is targeted by all the other branches.

tests/Database/Traits/EncryptableTest.php
tschallacka: Yes, but I think the encryptable traight might need to be modified, as laravel now supports encrypting attributes.

https://github.com/illuminate/database/blob/master/Eloquent/Concerns/HasAttributes.php#L131
https://github.com/illuminate/database/blob/master/Eloquent/Concerns/HasAttributes.php#L934

using something like

```php
<?php

namespace App;

use App\Casts\EncryptCast;
use Illuminate\Database\Eloquent\Model;

class User extends Model
{
    protected $casts = [
        'age' => EncryptCast::class,
    ];

    // ...
}
```
I had set the trait to leverage the use of the static encrypter variable https://github.com/wintercms/storm/pull/26/files#diff-b89596612b5ccf11dfa4953e4290a2723f7972adf31aa9d9ee47efdaf500f9c0R106 instead of the local variable.

But that should probably happen its own focussed pr?, I can revert it for now if you want.

LukeTowers: Yeah, let's do that. I don't feel comfortable mashing them together at the moment, that would be better reviewed in a separate PR with its own discussion.

src/Database/Pivot.php
tschallacka: So I assume this change can stay then, right? I just tried it with your composer update tip(you learn something every day) but it still throws a warning(not a fatal one luckily) So shall we keep it this way/mark it as resolved?

non fatal error

composer.json
tschallacka: I would keep this at 8 for now for the work that's occurring so that if the workflow comes that you can do a version release before they release Laravel 9, you can continue on and release 1.2 without being dependent on "in flux" code, and then target 9 in 1.3.

But that's up to you/the rest of the core team to decide, you have a better overview of the general development progress towards a new release 1.2 and if it will be released before/after laravel 9 hits the market.

tests/Database/Traits/EncryptableTest.php
tschallacka: It should be reset now back to the original implementation. The only change is the variable name due to the variable name conflict with the static variable in the parent class.

src/Foundation/Exception/Handler.php
tschallacka: Exception and Throwable imports are gone now, slashes are passed around like a grandma handing out candy on halloween.

src/Events/Dispatcher.php
tschallacka: Done, added a unit test for the dispatcher that tests all 3 types of event listeners it should accept.
I have no clue how to trigger a QueueClosure, as it requires a massive part of the framework to be booted up and I don't know how to get it working so the event bus works. I settled for testing that it accepts the argument without fault.

src/Database/Traits/Encryptable.php
tschallacka: This should be resolved now.

.github/workflows/tests.yml
tschallacka: Done.

composer.json
tschallacka: Done

composer.json
```diff
@@ -22,7 +22,7 @@
         }
     ],
     "require": {
-        "php": ">=7.2.9",
+        "php": ">=7.3",
```
LukeTowers:
```suggestion
        "php": ">=7.4",
```

composer.json
LukeTowers: We always only ever target LTS releases, so I think it's highly unlikely that we would target solely 8.x with 1.2

src/Database/Pivot.php
LukeTowers: Yes, the change can stay, was just providing historical context

src/Database/Traits/Encryptable.php
```diff
@@ -119,6 +119,6 @@ public function getEncrypter()
      */
     public function setEncrypter(\Illuminate\Contracts\Encryption\Encrypter $encrypter)
     {
-        $this->encrypter = $encrypter;
+        $this->encrypterInstance = $encrypter;
```
LukeTowers, line 122: This should probably also remain the same property that it was to not conflict with the new functionality in Laravel

src/Events/Dispatcher.php
```diff
@@ -30,8 +32,12 @@ class Dispatcher extends BaseDispatcher
...
     {
+        if ($events instanceof Closure || $events instanceof QueuedClosure) {
+            return parent::listen($events, $listener, $priority);
```
LukeTowers: I think we need to actually handle the processing of the closure events ourselves as Laravel does not include support for event priorities in their event dispatcher. Thoughts?

src/Foundation/Exception/Handler.php
LukeTowers: My apologies, we can still import the Throwable so we don't have to have slashes everywhere in the file, just wanted the exception import removed because as far as I could tell the exception class wasn't being used anywhere in the class.

tests/fixtures/events/EventTest.php
```diff
@@ -0,0 +1,5 @@
+<?php 
+
+class EventTest {
+    
+}
```
LukeTowers:
```suggestion
}

```

LukeTowers (changes requested):

tests/Database/Traits/EncryptableTest.php
LukeTowers: Ah, gotcha, that makes my other comment redundant. Thanks!

src/Foundation/Exception/Handler.php
tschallacka: It's used here. A lone usage. Do you want me to return the imports so the slashes can go? https://github.com/wintercms/storm/blob/e81f03b9b3450a6dc824c6bc6647544ddedce851/src/Foundation/Exception/Handler.php#L181

src/Events/Dispatcher.php
tschallacka: That's a lot of rewriting I would gather, as everything seems to be string orientied and would neeed to be refactored to account for the closures and the QueuedClosures. 
The latter are nightmarish for testing, as they require the event bus to be up and running, and I have tried and failed getting that working in a unit test for 2.5 hours. Google wanst much help either.
But then again, I'm relatively unexperienced bootstrapping and mocking core laravel components.

I could take a jab at it if I could test those queuedClosures working and triggereng the closure on event.

src/Foundation/Exception/Handler.php
LukeTowers: Is there anything in the base class that would suggest we can't just make that use throwable as well?

src/Events/Dispatcher.php
LukeTowers: I suppose that's fine for now then, I don't really know anything about closure based event listeners in the first place. Feel free to toss a TODO comment in there with references to further information and we can take another look at it closer to 1.2

tschallacka: The normal closures are pretty straight forward. They just run when the condition is triggered when the event is fired with the appropriate type hint(think `function (HoppingBunnyEvent $event) {}`,   
Not that different from the string based event handlers, except you can completely bypass the appropriate string finding and use your ide's autocomplete to find the correct class. 

It's the queued closure that's tricky. It get scheduled as a job, which then presumably gets run when `php artisan queue:work --queue=high,default` is called.

So all the parts that are involved with scheduling tasks/jobs need to be booted for a proper test so it can ben known it works and is accepted properly and registered properly etc...  Especialy when we want to add priorities that sort the jobs.

Then there are two queues high, low it seems, https://laravel.com/docs/8.x/queues#queue-priorities, need to research if those can define even more queues so we can implement a sort of priority queue where they can be sorted into, or simply just only accept high/low as parameters when it's a queued job.

composer.json
tschallacka: I changed the targetting to dev master. For tinker I needed to target dev-develop specifically to get it working. When they merge that in their master we can target dev-master there.
All unit tests on our side pass.

src/Events/Dispatcher.php
LukeTowers: I don't think we'd want to support priorities on queued event listeners, it sounds like it's more so for the use case of "hey, use this event being fired as a trigger that I should take care of a job later" rather than actually looking to do anything to filter or modify existing behavior in an event handler.

src/Foundation/Exception/Handler.php
tschallacka: I made some changes to the class, everything is now throwable oriented.

src/Events/Dispatcher.php
tschallacka: Okay, that makes it easy then. Queueables get pushed to parent, closure event listeners get priority treatment. I'll see what I can do :-)

tschallacka: So uhm, I did it, I looked at the laravel code with fresh eyes and saw how it could be easily implemented into our event handler queue without breaking stuff.

However, I did add a [parameter description](https://github.com/wintercms/storm/pull/26/commits/1cda0ebc75eec61ce1c0fa37c73f277370697a55#diff-1c29f9614d646d36e4f57a6a43cdcbf485f88eeae3a6fa929355f4224d536299R31) describing the duality of one parameter when a Closure type is passed.

Did I describe it correctly, it is this undesired behaviour and should they pass a null  and then priority?

so do you want:
```php
$dispatcher->listen(function(Event $foo){}, null, 2);
```
or do you want(what I've currently coded in)

```php
$dispatcher->listen(function(Event $foo){}, 2);
```

tschallacka:
@LukeTowers I think I've handled everything now for the migration, even that pesky QueuedClosure listener is now tested and priorities work for all the things. I just need feedback on https://github.com/wintercms/storm/pull/26#discussion_r627785448 and otherwise I think all is done for this library?

src/Events/Dispatcher.php
LukeTowers: The current solution you have looks fine. How does passing a Queued Closure work now?

tschallacka: A queuedevent  when initialized gets the closure that's passed to it assigend to the property closure.
Then for the event handler the method QueuedClosure::resolve() is called.
Resolve wraps the QueuedClosure->closure in a new closure which is returned. 
This resolved closure injects it into the queue scheduler when it is executed https://github.com/laravel/framework/blob/29b019774b746c90534465f1b949767707d49245/src/Illuminate/Events/QueuedClosure.php#L115
When the event class is triggered it is this resolved closure/job injector that is executed, to schedule the passed event closure.

For testing I mocked the QueuedClosure and overrode the resolve method, so it wouldn't construct an injector but just return the given closure.

mjauvin:
@tschallacka do you know if Closures are now serializable in Laravel 8/9 ?

ref. 
https://github.com/octobercms/library/pull/551/files
https://github.com/octobercms/library/pull/462

tschallacka:
@mjauvin a QueuedClosure is, https://github.com/laravel/framework/blob/29b019774b746c90534465f1b949767707d49245/src/Illuminate/Events/QueuedClosure.php#L6

mjauvin:
@tschallacka ok, but I am refering to regular closures provided as an event listener. If the Object which handle the event gets cached (serialized), any closure used as the event handler will currently cause an error when trying to serialize the object.

tschallacka:
Can you provide a unit test code that mimicks that?  I'll see if I can do something about that.

mjauvin:
@tschallacka you can try this test CMS page:
```
title = Closure Serialization Test Page 
url = /cache 
== 
function onInit()
{
    \Cms\Classes\Page::extend(function ($model) {
        $model->bindEvent('cms.page.display', function () {});
    });

    $page = new \Cms\Classes\Page([
        'fileName' => 'test-page',
        'title' => 'My Test Page',
        'url' => '/test-page',
    ]);

    Cache::put('my-cached-page-key', $page);
}
==
Hello World!
```

tschallacka:
What a strange place to extend a model. It should be extended in the boot method of the plugin...

I'll have a look at it though.

mjauvin:
You can apply this wherever you want, just a quick way to demonstrate

mjauvin:
You can also test this in artisan tinker like this (easiest way):

```
# php artisan tinker
Psy Shell v0.10.8 (PHP 8.0.5 — cli) by Justin Hileman
>>> class MyTest extends \Model
... {
... }
>>> 
>>> MyTest::extend(function ($model) {
...     $model->bindEvent('cms.page.display', function () {});
... });
=> null
>>> $obj = new MyTest();
=> MyTest {#3787}
>>> Cache::put('my-obj-key', $obj);
Exception with message 'Serialization of 'Closure' is not allowed'
>>> 
```
If you comment out `$model->bindEvent('cms.page.display', function () {});` you won't get the error

tschallacka:
@mjauvin that's a good example I can put in a unit test. Thanks.

tschallacka:
@mjauvin I've added some serialisation protection. Referenced variable connections via `use (&$var)` get severed during serialisation, but `use ($var)` persists.

@LukeTowers I've added a lot in the latest commits, see the commit message for a broad overview, if you have any questions just ask.

I still need to check Dispatcher for closures. Are there any other classes that work via closures that need to be serializeable?

src/Extension/ExtensionTrait.php
```diff
@@ -34,6 +36,9 @@ public function extensionApplyInitCallbacks()
         foreach ($classes as $class) {
             if (isset(self::$extensionCallbacks[$class]) && is_array(self::$extensionCallbacks[$class])) {
                 foreach (self::$extensionCallbacks[$class] as $callback) {
+                    if ($callback instanceof SerializableClosure) {
+                        $callback = $callback->getClosure();
```
tschallacka: I need to modify this to Serialisation::unwrapClosure

src/Extension/ExtensionTrait.php
```diff
@@ -54,28 +59,30 @@ public static function extensionExtendCallback($callback)
         ) {
             self::$extensionCallbacks[$class] = [];
         }
-
+        if ($callback instanceof \Closure && !($callback instanceof SerializableClosure)) {
```
tschallacka: I need to modify this to Serialisation helper call

src/Extension/ExtensionConstants.php
```diff
@@ -0,0 +1,15 @@
...
+ * @package Winter\Storm\Extension
+ */
+class ExtensionConstants
+{
+    const EXTENSIONS = 'extensions';
+    const METHODS = 'methods';
+    const FIELDS = 'fields';
+    const DYNAMIC_METHODS = 'dynamicMethods';
+    const DYNAMIC_PROPERTIES = 'dynamicProperties';
+}
```
LukeTowers: I don't think this is worth the added verbosity to the code. What are your thoughts?

tschallacka: Magic strings should be avoided when possible IMO. A magic string is subject to human error, and requires a lot of copy pasting, will not be error checked.

For some of the other reasons check

https://softwareengineering.stackexchange.com/questions/365339/what-is-wrong-with-magic-strings

tschallacka: Now that I think about it, it might even be better to replace the array access used in the extender with management classes so it becomes more self documenting instead of the nested array properties? Your thoughts?

LukeTowers: I would have to disagree in this specific case. The Extendable functionality is already quite complex, so adding more complexity to it doesn't make it better. Also, the magic strings are for an internal API that we do not want to be encouraging developers to be touching at all, so personally I'm fine with leaving them as magic strings because it's not like it's a public API that should be interacted with by developers on a regular basis.

tschallacka: It's changed back now.

src/Extension/ExtendableTrait.php
```diff
@@ -20,10 +22,10 @@ trait ExtendableTrait
...
      */
     protected $extensionData = [
-        'extensions'        => [],
-        'methods'           => [],
-        'dynamicMethods'    => [],
-        'dynamicProperties' => []
+        'extensions'         => [],
+        'methods'            => [],
+        'dynamicMethods'     => [],
+        'dynamicProperties'  => []
```
```suggestion
        'extensions'        => [],
        'methods'           => [],
        'dynamicMethods'    => [],
        'dynamicProperties' => []
```

tschallacka:
Should I squash the commits? And apologies for forgetting to run the code quality check.

LukeTowers:
Don't worry about squashing the commits, we take care of the history when we merge.

LukeTowers:
Thanks bot for doing the job you're meant to do :) 

I still need to take another look at this, but @bennothommo have you had a look yet?

bennothommo:
@LukeTowers not really - was going to take a look when Laravel 9 was out.

izdrail:
ok luke ,
 where do you think is the  problem ? regading this PR ?

LukeTowers:
@lzomedia our main list for tracking progress on this is available here: https://github.com/wintercms/winter/pull/148. This PR is good enough to merge into our WIP branch at the moment though. Thanks @tschallacka for your work on this so far!

PR #27, opened 2021-04-28
Namespace alias support in extendable trait

jaxwilko:
This PR adds support for `ExtendableTrait::isClassExtendedWith` to use namespace aliases registered within the `ClassLoader`. This fixes the issue @mjauvin found with `Winter.Translate` not working correctly with RainLab implements.

To test (provided by @mjauvin):

```
1. install rainlab.user, rainlab.location, rainlab.translate, offline.mall
2. run the migrations
3. in the backend, Settings->Translate->Manage languages
4. add a second language and enable it
5. go to the Catalogue menu and click `Create` button to create a new Product.
6. notice the language selector in the Product name field (among others).
```
To trigger the issue:
```
1. install winter.translate
2. run migrations
3. go back to Catalogue
4. click `Create` button
5. notice the Product name field is missing the language selector
```

The issue was caused by this code:

image

Within: `plugins/winter/translate/classes/EventRegistry.php`


jaxwilko:
I'll fix the tests + add new ones in the morning. I also want to add some extra checks in the code to handle the namespace splitting

LukeTowers:
Are there any risks related to timing concerns about trying to call these methods before aliases have been registered or before the application is available.

jaxwilko:
There could be with things like plugin `boot` functions, definitely worth testing for before merging

src/Extension/ExtendableTrait.php
```diff
@@ -225,7 +226,50 @@ public function extendClassWith($extensionName)
...
+            if ($aliases = $this->getClassLoader()->getNamespaceAliases($namespace)) {
+                foreach ($aliases as $alias) {
+                    if (isset($this->extensionData['extensions'][str_replace($namespace, $alias, $name)])) {
+                        return true;
+                    }
+                }
+            }
+        }
+
+        return false;
```
LukeTowers: Can't we just do 
```php
return isset($this->extensionData['extensions'][$name]) || isset($this->extensionData['extensions'][$classLoader->getAlias($name)])
```

jaxwilko: Not really, the aliases returned are an array as you can have multiple aliases registered pointing back to one original. This also differs from class aliasing as it's an alias of a segment of the namespace, so we need to break down the namespace of `$name` and check if any part of it has an alias that is being used for the implement. Sorry if that isn't clear, it's 4am :) lol

LukeTowers: So is the extensionData array populated with the class names as they were when they were registered instead of the real class name of the extension? If that's the case, could we not resolve the class alias using getAlias() when the extension is first registered, and then always resolve aliases when checking for existence?

jaxwilko: Yes to the first question, not sure about the second, I don't think there was an alias set for the implement class specifically as it was covered by the namespace alias registered during the `PluginManager` init. So we have to more or less guess at the namespace. I could take a look tomorrow though and see what we can do

LukeTowers: getAlias() resolves namespace aliases FYI. Tomorrow sounds good though, get some sleep already! 💤

jaxwilko: image

So `getAlias` will resolve an alias to the original, but what we actually want to check is if a class has an alias and if that alias is being used. 

I.e.
- Find all aliases for the namespace being used
- Check each to see if it is being implmented

LukeTowers: > what we actually want to check is if a class has an alias and if that alias is being used.
Why do we need to check that?

jaxwilko: So for instance, if the user has installed:
- Winter.Translate
- Offline.Mall

Mall has a `Product` model that implement `RainLab.Translate.Behaviors.TranslatableModel`. Which is handled by the class loader when the model gets extended, but `ExtendableTrait::isClassExtendedWith()` will return false if you pass it `Winter.Translate.Behaviors.TranslatableModel` because it's not in the `extensionData` array.

So when calling `isClassExtendedWith()` we want to check if the object has been extended with an alias of a original class.

jaxwilko: Another option might be making changes here:

image

And modifying `$extensionName` to be the actual name of object created.

No idea what this'll break tho lol

LukeTowers:
We should probably also do some performance testing on this with a heavy load because it's so level within the application to determine if it makes sense to only touch the class loader as little as possible (i.e. for isClassExtendedWith & getClassExtension we could check to see if the non-alias resolve class name exists before asking for the alias; but I don't know if that will actually make a positive improvement to performance or not without testing it)

src/Extension/ExtendableTrait.php
```diff
@@ -525,4 +517,18 @@ public static function extendableCallStatic($name, $params = null)
             $name
         ));
     }
+
+    /**
+     * Gets the class loader
+     *
+     * @return ClassLoader|null
+     */
+    protected function getClassLoader(): ?ClassLoader
```
LukeTowers: @bennothommo it was named extensionGetClassLoader() to avoid any potential conflicts since this is basically **the** base class for everything in October

bennothommo: Ah I see - I'll rename it back.

jaxwilko:
Just done some testing using the Winter translate behaviors and everything seems okay :)

mjauvin:
Any more concerns for merging this?

bennothommo:
Replaced with #38.

PR #29, opened 2021-04-29
Improved the unique validation rule test cases

LukeTowers:
Resolves some inconsistencies noticed by @Flynsarmy.

@bennothommo @Flynsarmy if you guys could do a sanity check on this for me please that would be great.

### DIFF
--- a/src/Database/Traits/Validation.php
+++ b/src/Database/Traits/Validation.php
@@ -1,13 +1,14 @@
 <?php namespace Winter\Storm\Database\Traits;
 
 use App;
-use Exception;
-use Input;
 use Lang;
-use Illuminate\Contracts\Validation\Rule;
-use Winter\Storm\Support\Facades\Validator;
+use Input;
+use Exception;
 use Illuminate\Support\MessageBag;
+use Illuminate\Contracts\Validation\Rule;
+use Winter\Storm\Support\Str;
 use Winter\Storm\Database\ModelException;
+use Winter\Storm\Support\Facades\Validator;
 
 trait Validation
 {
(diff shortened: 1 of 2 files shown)


src/Database/Traits/Validation.php
```diff
@@ -435,31 +436,42 @@ protected function processRuleFieldNames($rules)
...
+        $params = ["unique:{$connection}.{$table}", $column, $ignoreValue, $ignoreColumn];
 
-        if ($whereColumn) {
+        if ($whereColumn && $whereValue) {
             $params[] = $whereColumn;
-        }
-
-        if ($whereValue) {
             $params[] = $whereValue;
         }
```
bennothommo: @LukeTowers would this break conditions that don't have a value - not sure if it's possible, but I wonder if this method previously supported things like "ISNULL(deleted_at)", which technically does not need a value to be used as a condition.

LukeTowers: @bennothommo good catch, looks like that is supported in the Unique rule object at least: https://github.com/laravel/framework/blob/8.x/src/Illuminate/Validation/Rules/DatabaseRule.php#L93-L95. Mind adding a test for it and pushing the fix?

src/Database/Traits/Validation.php
```diff
@@ -435,31 +436,42 @@ protected function processRuleFieldNames($rules)
...
             $whereValue
         ) = array_pad(explode(',', $definition, 6), 6, null);
 
-        $table = 'unique:' . $this->getConnectionName()  . '.' . $this->getTable();
-        $column = $column ?: $fieldName;
-        $key = $keyName ? $this->$keyName : $this->getKey();
-        if (is_null($key)) {
-            $key = 'NULL';
+        // Remove unique or unique: from the table name
+        $table = ltrim(ltrim($table, 'unique'), ':');
```
LukeTowers:
```suggestion
        $table = Str::after(Str::after($table, 'unique'), ':');
```

LukeTowers:
@bennothommo is this good to go?

bennothommo:
@LukeTowers yep, I believe so. :)

PR #30, opened 2021-05-07
Remove addNamespaceAliases() method

mjauvin:
Related to https://github.com/wintercms/winter/pull/159#issuecomment-834762555


LukeTowers:
Please remove all of the lazy class aliasing functionality that was added, it doesn't work fully so there's no reason to keep it.

bennothommo:
This won't be needed if we go ahead with #27 and #31.

PR #31, opened 2021-05-11
Add reverse aliases for lazy-loaded classes.

bennothommo:
This should resolve an issue where a replacing plugin's class won't report as the replaced plugin's class in `instanceof` calls and the like, by aliasing a loaded plugin class that is being aliased to and reserving the names of the replaced classes.

We also track which aliases have been loaded and/or reversed to prevent double-initialisation.


mjauvin:
What's the status on this? Are we confident enough to merge?

bennothommo:
Replaced with #38.

PR #32, opened 2021-05-14
[WIP] Added reverse alias support in ExtendedTrait

jaxwilko:
This is kind of a POC for an approach to the problem of having `App` unavaliable during tests.

### DIFF
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -1,9 +1,11 @@
 <?php namespace Winter\Storm\Extension;
 
+use App;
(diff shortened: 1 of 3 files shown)


LukeTowers:
What's the difference between this and the changes made in the other PR?

jaxwilko:
@LukeTowers probably not much at this point. The idea was to more or less create a fix based on @bennothommo's branch using the new `getReverseAlias` method without any of the historic changes we made in the other PR. It was just meant to be a POC as I'm still not 100% happy with it, this one can probably be closed and the other PR worked on :)

PR #34, opened 2021-05-21
Reduce the calls to PathResolver::resolve for speed

jaxwilko:
During the Halcyon query for CMS pages, `FileDatasource::makeDirectoryPath` is called to generate the record file path. 

Within that method we call `PathResolver::within` to validate that the file found exsits within the base path of the model. 

Internally that method calls resolve on each arg passed, then returns `starts_with` which is fine, the issue is in the resolve method.

When returning, the `$path` var is resolved once again.

This PR reduces the calls to `PathResolver::resolve` by remembering the base path once initially resolved, then handling the `$path` var's resolution within the method scope so it can be tested and returned without needing to be resolved for a second time.

The following are messurements of the time it has taken the `while` loop in `FileDatasource::select` to execute (in seconds).

All testing has been done with 1000 CMS pages.

Timeing changes

### DIFF
--- a/src/Halcyon/Datasource/FileDatasource.php
+++ b/src/Halcyon/Datasource/FileDatasource.php
@@ -336,7 +336,7 @@ class FileDatasource extends Datasource implements DatasourceInterface
         $path = !empty($relativePath) ? $base . '/' . $relativePath : $base;
 
         // Resolve paths with base lookup for performance
-        $base = $this->resolvedBasePaths[$base] ?? $this->resolvedBasePaths[$base] = PathResolver::resolve($base);
+        $base = $this->resolvedBasePaths[$base] ?? ($this->resolvedBasePaths[$base] = PathResolver::resolve($base));
         $path = PathResolver::resolve($path);
 
         // Limit paths to those under the configured basePath + directory combo
--- a/src/Halcyon/Datasource/FileDatasource.php
+++ b/src/Halcyon/Datasource/FileDatasource.php
@@ -32,6 +32,13 @@ class FileDatasource extends Datasource implements DatasourceInterface
     protected $files;
 
     /**
+     * Resolved path map.
+     *
+     * @var array
+     */
+    protected $pathResolverMap = [];
+
+    /**
      * Create a new datasource instance.
      *
      * @param string $basePath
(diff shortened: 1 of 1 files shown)


src/Halcyon/Datasource/FileDatasource.php
```diff
@@ -325,19 +332,19 @@ protected function validateDirectoryForSave(string $dirName, string $fileName, s
...
     protected function makeDirectoryPath($dirName, $relativePath = '')
     {
-        $base = $this->basePath . '/' . $dirName;
-        if (!empty($relativePath)) {
-            $path = $base . '/' . $relativePath;
-        } else {
-            $path = $base;
-        }
+        $base = $this->basePath . DIRECTORY_SEPARATOR . $dirName;
+        $path = !empty($relativePath) ? $base . DIRECTORY_SEPARATOR . $relativePath : $base;
```
bennothommo: @jaxwilko I know your way is better, but given this class still uses forward-slashes throughout, `DIRECTORY_SEPARATOR` might give some inconsistent results.

```suggestion
        $base = $this->basePath . '/' . $dirName;
        $path = !empty($relativePath) ? $base . '/' . $relativePath : $base;
```

src/Halcyon/Datasource/FileDatasource.php
```diff
@@ -31,6 +31,13 @@ class FileDatasource extends Datasource implements DatasourceInterface
      */
     protected $files;
 
+    /**
+     * Resolved path map.
+     *
+     * @var array
+     */
+    protected $pathResolverMap = [];
```
bennothommo: @jaxwilko Could we change this to `$resolvedBasePaths`? More clearly indicates what's being stored here. You'll need to change the references further on in the class too.

```suggestion
     * A list of resolved base paths.
     *
     * @var array
     */
    protected $resolvedBasePaths = [];
```

bennothommo (changes requested):

src/Halcyon/Datasource/FileDatasource.php
jaxwilko: Fair enough, i did wonder why the const wasn't being used, happy to change it back, was just doing some cleanup while i was there

src/Halcyon/Datasource/FileDatasource.php
jaxwilko: Happy with this change too, I've got a bunch of debug code in there atm for another PR I'm looking at but I'll try and get that one sorted and make this change later today :)

jaxwilko: This has been changed in https://github.com/wintercms/storm/pull/34/commits/faece692a56b3544d4d0ee4318b865dd2eef7606

src/Halcyon/Datasource/FileDatasource.php
```diff
@@ -326,18 +333,18 @@ protected function validateDirectoryForSave(string $dirName, string $fileName, s
...
         $base = $this->basePath . '/' . $dirName;
-        if (!empty($relativePath)) {
-            $path = $base . '/' . $relativePath;
-        } else {
-            $path = $base;
-        }
+        $path = !empty($relativePath) ? $base . '/' . $relativePath : $base;
+
+        // Resolve paths with base lookup for performance
+        $base = $this->resolvedBasePaths[$base] ?? $this->resolvedBasePaths[$base] = PathResolver::resolve($base);
```
LukeTowers:
```suggestion
        $base = $this->resolvedBasePaths[$base] ?? ($this->resolvedBasePaths[$base] = PathResolver::resolve($base));
```
For clarity

LukeTowers:
@jaxwilko do we have unit tests covering this method?

jaxwilko:
@LukeTowers I can't find any direct tests, however it would be indirectly tested via the other tests. Want me to look at writing some additional tests?

LukeTowers:
If you could that would be great. Try to break the path validation as much as possible

jaxwilko:
@LukeTowers shall do :)

jaxwilko:
@LukeTowers I've added the tests into the `CmsObjectTest` in the winter repo. It's included in this PR: https://github.com/wintercms/winter/pull/192/ which is independed of this PR, however the test will cover the before and after state of this PR as the output should be the same. Tests added in this commit: https://github.com/wintercms/winter/pull/192/commits/ccda179642cf3bbe3fcabf7e4bde7b39bafcfee2

bennothommo:
I'm happy with this as is - the test cases in the library already have some stringent checks for `PathResolver`, and they're still passing with these changes.

PR #35, opened 2021-06-07
Improve stubs adding missing translation keys

RomainMazB:
Leaving this as draft: I need to create some corresponding keys into the modules files and do some tests.

Also, using translation key, a little regression is needed: we can't underline the first letter which corresponds to the saving shortcut on update [here](https://github.com/wintercms/storm/pull/35/files#diff-fb3312381ad7f60ad56287d6ebd14cc9c9a7f9d7146b1363a93dd95b2e9b74feL25) because even if it would be possible to use css styling to underline the first letter, it won't be the good one in some language...


mjauvin:
There was a similar PR on October from Alxy, it never got merged because it added the localization strings without the  --include-translations argument...

see here:

https://github.com/octobercms/library/pull/487

RomainMazB:
I didn't know! Thanks!

I definitely get the @alxy work here and finish it before getting this PR out of draft status.

RomainMazB:
@wintercms/maintainers I've added a note into the doc to generate the controller's file using this option.
This PR is ok for me, I need feedback.
The diffcheckers for the generated view files to easily understand what I'm trying to achieve:
- [_list_toolbar.htm](https://www.diffchecker.com/KyugC9kW)
- [create.htm](https://www.diffchecker.com/jlJrFcb4)
- [preview.htm](https://www.diffchecker.com/BpIK4j9Q)
- [update.htm](https://www.diffchecker.com/7CrHetZO)

The goal of this one is to ease the build of a multi-languages plugin without the need to go to all the generated view files to manually operate those changes.
The generated files now including core's translations keys, the controller's views are out of the box mostly compatible with any maintained language.

I agree that it's not perfect out the box: example if I'm using this in a french app into a model named "car", the delete confirmation, will be like "Êtes-vous sûr de vouloir supprimer ce car?", which _-in french-_ is not what the plugin's developer mean (car = bus).

But still, this eases the developer work as he only needs to translate the model names and use nested translation like so:
```php
trans('backend::lang.form.confirm_delete_name', ['name' => trans('acme.rental::lang.models.car')])
```

> Note: it would be great if the trans method was including a fallback as parameter to avoid this: "Are you sure you want to delete acme.rental::lang.models.car?"

LukeTowers:
Replaced by https://github.com/wintercms/winter/pull/486

PR #37, opened 2021-06-08
Add create:settings command

RomainMazB:
Discussed here: https://github.com/wintercms/winter/discussions/201


src/Scaffold/Console/model/settings-model.stub
```diff
@@ -0,0 +1,18 @@
...
+{
+    public $table = '{{lower_author}}_{{lower_plugin}}_{{snake_plural_name}}';
+    use \Winter\Storm\Database\Traits\Validation;
+
+    public $implement = ['System.Behaviors.SettingsModel'];
+
+    public $settingsCode = '{{lower_author}}_{{lower_plugin}}_{{lower_name}}';
+
+    public $settingsFields = 'fields.yaml';
+}
```
jaxwilko: Would be nice to generate the model with comments. Also the table property doesn't need to be defined for settings models but we do need to define rules if the validation trait is in use.

```suggestion
<?php namespace {{studly_author}}\{{studly_plugin}}\Models;

use Model;

/**
 * {{name}} Model
 */
class {{studly_name}} extends Model
{
    use \Winter\Storm\Database\Traits\Validation;

    /**
     * @var array Behaviors implemented by this model.
     */
    public $implement = ['System.Behaviors.SettingsModel'];

    /**
     * @var string Unique code
     */
    public $settingsCode = '{{lower_author}}_{{lower_plugin}}_{{lower_name}}';

    /**
     * @var mixed Settings form field definitions
     */
    public $settingsFields = 'fields.yaml';

    /**
     * @var array Validation rules
     */
    public $rules = [];
}

```

LukeTowers:
Side note, would it instead be better to have `create:settings` rather than a flag on `create:model`?

RomainMazB:
> Side note, would it instead be better to have `create:settings` rather than a flag on `create:model`?

You are the one who determine that, just let me know.

My thoughts on that:
Upside: less verbose, stubs directory structure cleaner
Downside: less intuitive for beginners: we create a settings but it's created into the models directory

LukeTowers:
@RomainMazB I think I prefer `create:settings Author.Plugin ?OptionalSettingsClassOverride` rather than `create:model Author.Plugin SettingsClass --settings`. @wintercms/maintainers any thoughts?

LukeTowers:
@RomainMazB any thoughts on my comment?

LukeTowers:
@RomainMazB let's go with `create:settings` instead, and if it's possible (and not already done so) we should have a success message after the files are created that lists the files that were created as a result of the command so that it's very clear where their files were created and what they are.

RomainMazB:
> @RomainMazB let's go with `create:settings` instead, and if it's possible (and not already done so) we should have a success message after the files are created that lists the files that were created as a result of the command so that it's very clear where their files were created and what they are.

`create:settings` is now ok.
To list the files created during all the `create:`, we should improve the [`GeneratorCommand` class](https://github.com/wintercms/storm/blob/6c2edbb9e613fcd0c1fa70f806afda82265d0c7c/src/Scaffold/GeneratorCommand.php#L59-L66) which only displays a successful message atm.

I'm on it :)

src/Scaffold/Console/settings/fields.stub
```diff
@@ -0,0 +1,7 @@
+# ===================================
+#  Form Field Definitions
+# ===================================
+
+fields:
+    settings_option:
+        label: This is a sample settings used by {{author}}.{{plugin}}
```
LukeTowers:
```suggestion
        label: This is a sample settings field used by {{author}}.{{plugin}}
```

src/Scaffold/Console/CreateSettings.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * The type of class being generated.
+     *
+     * @var string
+     */
+    protected $type = 'Settings Model';
+
+    /**
+     * A mapping of stub to generated file.
```
LukeTowers:
```suggestion
     * A mapping of stubs to generated files.
```

PR #38, opened 2021-06-14
Improved aliasing and plugin replacement

bennothommo:
Combines the following PRs for better testing:

- https://github.com/wintercms/storm/pull/27 - Namespace alias support in extendable trait
- https://github.com/wintercms/storm/pull/31 - Add reverse aliases for lazy-loaded classes


phpunit.xml.dist
```diff
@@ -0,0 +1,23 @@
...
+        <include>
+            <directory suffix=".php">./src/</directory>
+        </include>
+    </coverage>
+    <testsuites>
+        <testsuite name="Storm Library Test Suite">
+            <directory>./tests</directory>
+        </testsuite>
+    </testsuites>
+</phpunit>
```
LukeTowers, line 23: Why was this added?

bennothommo: It's a testing config for PHPUnit. It was added so instead of running `./vendor/bin/phpunit ./tests`, we could just run `./vendor/bin/phpunit` instead, and also allows us to use a code coverage service in the future, if we so choose.

We have a similar config set up in the main repo as well (https://github.com/wintercms/winter/blob/develop/phpunit.xml), but this one allows us to run the library tests in isolation.

LukeTowers: What's with the .dist extension?

bennothommo: `phpunit.xml.dist` is standard for reusable libraries (which Storm arguably is). It's a distributed "example" config which can be overridden locally by a `phpunit.xml`, but will be used if `phpunit.xml` doesn't exist.

mjauvin:
I've tested the following:

- `instanceof` works for classes of replaced plugins.
e.g. This works if Winter\Blog is installed and replaces `RainLab\Blog`:
```
   $model = new \Winter\Blog\Models\Post;
   $model instanceof RainLab\Blog\Models\Post; // returns true
```
- public $implement = []; // works for classes of a replaced plugin.
e.g. This works if `Winter\Translate` is installed and replaces `RainLab\Translate`:
```
use Model;

class myClass extends Model
{
    public $implement = ['@RainLab.Translate.Behaviors.TranslatableModel'];
}
```

LukeTowers:
Were we able to do any performance testing on this @mjauvin or @jaxwilko? This is such a low level feature that I want to make sure it's rock solid

bennothommo:
@mjauvin one extra thing to test is what happens if both RainLab.Blog and Winter.Blog are installed - do the replacements work correctly, or is still a requirement to remove the original plugins first? In my testing, it seemed to be fine with both plugins installed, but just wanted to see if someone else can confirm the same.

mjauvin:
@bennothommo I tested with both Winter/RainLab Blog plugin installed and it's working fine. @jaxwilko's code still detects and warns that both plugins are installed, but the replacement seems to not care for the replaced plugin.

jaxwilko:
@mjauvin @bennothommo Last I left it, the desired functionality is that `Winter.Blog` would replace `Rainlab.Blog` by disabling `Rainlab.Blog` and showing a warning suggesting it's removal. This way the `Rainlab.Blog` plugin will stay in place but wont be registered and no code within it will be executed.

mjauvin:
So, yeah, this code is what makes this work now in `Support/ClassLoader::load()`:
```
        // If the class is already aliased, skip loading.
        if (in_array($class, $this->loadedAliases) || in_array($class, $this->reversedClasses)) {
            return true;
        }
```

src/Support/Testing/MocksClassLoader.php
```diff
@@ -0,0 +1,65 @@
...
+     * Registers a mock class loader in the autoload stack.
+     *
+     * @param string $basePath
+     * @param string $manifestPath
+     * @return void
+     */
+    protected function registerMockClassLoader($basePath = null, $manifestPath = null)
+    {
+        if (is_null($basePath)) {
+            $basePath = dirname(dirname(__DIR__)) . '/fixtures/classes';
```
RomainMazB, line 26: Doesn't `dirname(__DIR__, 2)` looks more readable?

src/Support/Testing/MocksClassLoader.php
```diff
@@ -0,0 +1,65 @@
...
+        }
+        if (is_null($manifestPath)) {
+            $manifestPath = dirname(dirname(__DIR__)) . '/fixtures/classes/classes.php';
```
RomainMazB, line 29: Same here

RomainMazB (commented):
Just a little comment.

src/Support/Testing/MocksClassLoader.php
LukeTowers: I don't think I've seen the second argument be passed to dirname before in the codebase

src/Extension/ExtendableTrait.php
```diff
@@ -122,6 +124,21 @@ public static function clearExtendedClasses()
         self::$extendableCallbacks = [];
     }
 
+    /**
+     * Normalizes the provided extension name
```
```suggestion
     * Normalizes the provided extension name allowing for the ClassLoader to inject aliased classes
```

src/Extension/ExtendableTrait.php
```diff
@@ -122,6 +124,21 @@ public static function clearExtendedClasses()
...
+     *
+     * @param string $name
+     * @return string
+     */
+    protected function extensionNormalizeClassName(string $name): string
+    {
+        $name = str_replace('.', '\\', trim($name));
+        if (!is_null($this->extensionGetClassLoader()) && ($alias = $this->extensionGetClassLoader()->getAlias($name))) {
```
LukeTowers, line 141: @bennothommo can we cache this locally as `$this->extensionClassLoader`? Every call to `extensionNormalizeClassName` now results in two calls to `class_exists()` and `App::make()`

src/Support/ClassLoader.php
```diff
@@ -98,30 +112,59 @@ public function __construct(Filesystem $files, $basePath, $manifestPath)
...
+                }
             }
         }
 
-        if (!is_null($alias = $this->getAlias($class))) {
-            return class_alias($alias, $class);
+        if (!is_null($alias = $this->getAlias($class)) && !in_array($class, $this->reversedClasses)) {
+            $this->loadedAliases[] = $class;
+            class_alias($alias, $class);
+            return true;
```
LukeTowers, line 167: The docblock for this method is not accurate (currently `@return void`)

src/Support/ClassLoader.php
```diff
@@ -330,7 +427,10 @@ protected static function normalizeClass($class)
         $lowerClass = strtolower($directory) . DIRECTORY_SEPARATOR . $file . '.php';
         $upperClass = $directory . DIRECTORY_SEPARATOR . $file . '.php';
 
-        return [$lowerClass, $upperClass];
+        $lowerClassStudlyFile = strtolower($directory) . DIRECTORY_SEPARATOR . Str::studly($file) . '.php';
+        $upperClassStudlyFile = $directory . DIRECTORY_SEPARATOR . Str::studly($file) . '.php';
+
+        return [$lowerClass, $upperClass, $lowerClassStudlyFile, $upperClassStudlyFile];
```
LukeTowers, line 433: What was this change for?

bennothommo: @LukeTowers It was to catch some extra permutations of the filename - essentially allow for PSR-4 filenames. It was mainly for the Backend controller class, which searches a lower-case filename when determining the controller from the URL.

src/Support/ClassLoader.php
```diff
@@ -94,34 +108,63 @@ public function __construct(Filesystem $files, $basePath, $manifestPath)
      * Load the given class file.
      *
      * @param  string  $class
-     * @return void
+     * @return true|null
```
LukeTowers: `return class_alias()` returns true or false. Can we also indicate what the different return values mean?

LukeTowers:
```suggestion
     * @return bool|null
```

PR #40, opened 2021-06-30
[Enhancement] Add ArrayFile & EnvFile Parsing, replace ConfigWriter internal logic

jaxwilko:
This is a POC PR proposing a replacement for `ConfigWriter`. Currently, `ConfigWriter` is used to update php config files, for example during winter install.

In this PR I've added a new class `ConfigFile`, which can be used to modify any php config file without risking the same kind of corruption that can be caused using `ConfigWriter`. The reason for this improvement is due to the fundemental differences in approach, where `ConfigWriter` modifies the file contents as a string, `ConfigFile` uses `nikic/php-parser` to generate an ast of the config file and allows us to modify the code as php objects before "printing" them back out using the `WinterPrinter` (an exstension to `PhpParser\PrettyPrinter\Standard`).

### Example

Config file:
```php
<?php

return [
    'foo' => env('ENV_VALUE', 'bar'),
    'bar' => [
         'winter' => 'cms'
    ]
];
```
Sample usage:
```php
use Winter\Storm\Config\ConfigFile;

$config = ConfigFile::read('/path/to/file');
$config->set('foo', 'foo')
    ->set('bar.winter', 'is coming');
$config->write();
```
Will update the config file to:
```php
<?php

return [
    'foo' => env('ENV_VALUE', 'foo'),
    'bar' => [
         'winter' => 'is coming'
    ]
];
```

There are currently some issues when node types are swapped by setting a value that differs in type from the originally parsed value, but they should be simple to fix. I have modified the `winter:install` command to use `ConfigFile` over `ConfigWriter` and it works nicely, however it would be good to get some feed back on this core peice before opening a PR for the install command.

Please let me know any thoughts or suggestions you have :)


LukeTowers:
Oooh shiny :) Can't wait to review this :) 👀

mjauvin:
Would it be possible to add a comma after the last item of an array when writing the results to the file (to ease manual maintainability)?

jaxwilko:
@mjauvin ah, it already does, my example just doesn't show it.

For reference here is a config I just saved using `winter:install`:

image

Also new line padding on single line comments has now been fixed as of 524561f

bennothommo:
This is awesome @jaxwilko. Well done :)

jaxwilko:
The issues relating to updating a value from one type to another (i.e. int -> bool) has now be resolved. I added a load of code and so that we would:
- Check if note is same type as incoming value (i.e. currently string, updating to string).
- If match, update the node in place.
- If not match, create a new node of the correct type and replace the original node.

Rather than doing all that extra work, it's far simpiler if we just:
- Create a new node of the correct type and replace the original node

This solves the casting issue and cuts down on code / processing.

jaxwilko:
The following is documentation I'm writing now while it's fresh in my head.

### Documentation
`ConfigFile` provides the ability to update php config files in place. A side effect it can also be used to reformat a config file in the Winter style.

#### Init
To initalize the `ConfigFile` object, you can use the static `read()` method:
```php
$config = ConfigFile::read('/path/to/file');
```
Alternatively, you can directly initalize the object:
```php
$config = new ConfigFile($ast, $filePath = null, $printer = null);
```
However initalizing the object this way requires you to parse the config file manually using `PhpParser\Parser` and pass the result as the first argument.

#### Updating a value
To update a value in the config, call the `set()` method:
```php
$config->set('property.key', $value);
```
If you have multiple properties to update, you can also pass an array:
```php
$config->set([
    'property.key' => $value,
    'property.other_key' => $otherValue
]);
```

#### Saving
To retrieve the "printed" result as a string you can use the `render()` method:
```php
$string = $config->render();
```
If you set use the `ConfigFile::read()` method for initialization or set the `$filePath` property when initing the object, you can call the `write()` method to overwrite the original file:
```php
$config = ConfigFile::read('/path/to/config.php');
$config->write(); // saves to /path/to/config.php
```
Alternatively you can pass `write()` a file path to save to:
```php
$config = ConfigFile::read('/path/to/config.php');
$config->write('/path/to/config.new.php'); // saves to /path/to/config.new.php
```

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,182 @@
...
+     * @var PrettyPrinterAbstract|WinterPrinter|null
+     */
+    protected $printer = null;
+
+    /**
+     * Config constructor.
+     * @param $ast
+     * @param PrettyPrinterAbstract|null $printer
+     */
+    public function __construct($ast, string $file = null, PrettyPrinterAbstract $printer = null)
```
bennothommo: @jaxwilko given the check immediately below, would the `$ast` be an array?
```suggestion
    public function __construct(array $ast, string $file = null, PrettyPrinterAbstract $printer = null)
```

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        $content = file_get_contents($file);
+        $parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
+
+        try {
+            $ast = $parser->parse($content);
+        } catch (Error $e) {
+            // should add better handling
+            throw new Error($e);
```
bennothommo: @jaxwilko we should be throwing `ApplicationException`s by default if the failure is the result of a user input.
```suggestion
            throw new ApplicationException($e);
```

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,182 @@
...
+            foreach ($args[0] as $key => $value) {
+                $this->set($key, $value);
+            }
+
+            return $this;
+        }
+
+        if (count($args) !== 2 || !is_string($args[0])) {
+            throw new \InvalidArgumentException('invalid args passed');
+        }
```
bennothommo: @jaxwilko I think this section might be a bit too "magic" - we should probably be specific on the parameters in the signature at least.
```suggestion
    /**
     * @param string|array $key
     * @param string|null $value
     * @return $this
     */
    public function set($key, string $value = null): ConfigFile
    {
        if (is_array($key)) {
            foreach ($key as $name => $value) {
                $this->set($name, $value);
            }

            return $this;
        }

        if ($key && is_null($value)) {
            throw new ApplicationException('You must specify a value to set for the given key.');
        }
```

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,182 @@
...
+            return $this;
+        }
+
+        $target->value = $this->makeAstNode($valueType, $value);
+        return $this;
+    }
+
+    /**
+     * @param string $type
+     * @param $value
```
bennothommo:
```suggestion
     * @param mixed $value
```

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,182 @@
...
+
+                return $item;
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * @param string|null $filePath
```
bennothommo:
```suggestion
     * @param string|null $filePath
     * @return void
```

bennothommo (dismissed):
In addition to the below changes, could you provide some details in the docblocks for the main class methods? I assume these were auto-generated through PHPStorm, but don't really describe *what* the methods are doing.

src/Config/ConfigFile.php
jaxwilko: I'll have to check later but I'm pretty sure the parser returns mixed results, Ideally we want to take whatever we're given and then validate it inside the constructor to ensure it's a config style php file

jaxwilko:
@bennothommo I'll add some descriptive comments when I find my laptop :)

mjauvin:
> @bennothommo I'll add some descriptive comments when I find my laptop :)

You didn't throw it out the window yesterday, did you? :rofl:

src/Config/ConfigFile.php
jaxwilko: I double checked and array should be fine, I've added this in c09718739344ec6544dc690fa48d11de85b52a78

bennothommo:
@jaxwilko would you mind making sure that the scenario being fixed by #43 is covered by this rewrite? If you need to add any changes or the test cases from #43, make sure to `Co-Authored by` credit @ericp-mrel in the final merge :).

jaxwilko:
As mentioned here: https://github.com/wintercms/storm/pull/43, the `EnvFile` class has now been added.

I have some concerns with the approach however as it loses comments at the moment which could be very annoying depending on peoples work flows.

I have some ideas to fix this which I shall be looking into.

LukeTowers:
@jaxwilko what's the status on this?

jaxwilko:
@LukeTowers I've made a start converting the usages from old to new but have been really busy with some contract work the last few weeks. I'll try and finish off what i've got or push it onto a branch on the winter repo so it can be picked up by somebody else if they have time :)

LukeTowers:
Example syntax for localization support for scaffolding:

```php
protected function processVars($vars)
    {
        $vars = parent::processVars($vars);
        
        $varsToLangKeys = [
            'title_plural_name' => "{{author}}.{{plugin}}::lang.models.{{model}}.label_plural",
            'title_singular_name' => "{{author}}.{{plugin}}::lang.models.{{model}}.label",
        ];
        
        $langFilePath = plugins_path("{{author}}/{{plugin}}/lang/" . \App::getLocale() . "/lang.php");
        
        // This will load an existing file or create a new one if required
        $langFile = \Winter\Storm\Parse\ArrayFile::load($langFilePath);
        
        foreach ($varsToLangKeys as $varKey => $langKey) {
            $langFile->set(\Str::after($langKey, '::lang.'), $vars[$varKey]);
            $vars[$varKey] = $langKey;
        }
        
        $langFile->write();
        // Maybe $langFile->close(); if required.
    }
```

src/Config/ConfigConst.php
```diff
@@ -0,0 +1,33 @@
+<?php namespace Winter\Storm\Config;
+
+/**
+ * Class ConfigConst
+ * @package Winter\Storm\Config
+ *
+ * This class is for use with ConfigFile as a method to inject a constant into a config file
+ */
+class ConfigConst
```
LukeTowers: Let's go with `Winter\Storm\Parse\PHP\PHPConstant`

src/Config/ConfigFileInterface.php
```diff
@@ -0,0 +1,43 @@
...
+
+// Returns are commented out as improved type variance is not supported until php 7.4
+// Ref: https://stitcher.io/blog/new-in-php-74#improved-type-variance-rfc
+// TODO: enable return types once support for <7.4 is dropped
+
+/**
+ * Interface ConfigFileInterface
+ * @package Winter\Storm\Config
+ */
+interface ConfigFileInterface
```
LukeTowers: Let's go with `Winter\Storm\Parse\Contracts\FileInterface`

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+use PhpParser\Node\Stmt;
+use PhpParser\ParserFactory;
+use PhpParser\PrettyPrinterAbstract;
+use Winter\Storm\Exception\SystemException;
+
+/**
+ * Class ConfigFile
+ * @package Winter\Storm\Config
+ */
+class ConfigFile implements ConfigFileInterface
```
LukeTowers: Let's go with `Winter\Storm\Parse\PHP\ArrayFile`

src/Config/ConfigFileInterface.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    /**
+     * Return a new instance of `ConfigFileInterface` ready for modification of the file.
+     *
+     * @param string $file
+     * @return ConfigFile|null
+     */
+    public static function read(string $file); //: ?ConfigFileInterface;
```
LukeTowers:
```suggestion
    public static function open(string $filePath); //: ?ConfigFileInterface;
```
Matches the PHP core of `fopen()` `fwrite()` better

src/Config/ConfigFunction.php
```diff
@@ -0,0 +1,49 @@
+<?php namespace Winter\Storm\Config;
+
+/**
+ * Class ConfigFunction
+ * @package Winter\Storm\Config
+ *
+ * This class is for use with ConfigFile as a method to inject a function call into a config file
+ */
+class ConfigFunction
```
LukeTowers: Let's go with `Winter\Storm\Parse\PHP\PHPFunction`

src/Config/EnvFile.php
```diff
@@ -0,0 +1,268 @@
...
+
+use Winter\Storm\Config\ConfigFileInterface;
+use Dotenv\Environment\DotenvFactory;
+use Dotenv\Loader;
+
+/**
+ * Class EnvFile
+ * @package Winter\Storm\Config
+ */
+class EnvFile implements ConfigFileInterface
```
LukeTowers: Let's go with `Winter\Storm\Parse\EnvFile`

src/Config/WinterPrinter.php
```diff
@@ -0,0 +1,51 @@
+<?php namespace Winter\Storm\Config;
+
+use PhpParser\Node\Expr\ArrayItem;
+use PhpParser\PrettyPrinter\Standard;
+
+/**
+ * Class WinterPrinter
+ * @package Winter\Storm\Config
+ */
+class WinterPrinter extends Standard
```
LukeTowers: Let's go with `Winter\Storm\Parse\PHP\ArrayPrinter`

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+     */
+    protected $ast = null;
+    /**
+     * @var string|null Source config file
+     */
+    protected $file = null;
+    /**
+     * @var PrettyPrinterAbstract|WinterPrinter|null Printer used to define output syntax
+     */
+    protected $printer = null;
```
LukeTowers:
```suggestion
    /**
     * @var Stmt[]|null Abstract syntax tree produced by `PhpParser`
     */
    protected $ast = null;

    /**
     * @var string|null Path to the file 
     */
    protected $filePath = null;

    /**
     * @var PrettyPrinterAbstract|WinterPrinter|null Printer used to define output syntax
     */
    protected $printer = null;
```

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+
+    /**
+     * ConfigFile constructor.
+     *
+     * @param Stmt[]|null $ast
+     * @param string $file
+     * @param PrettyPrinterAbstract|null $printer
+     */
+    public function __construct(array $ast, string $file = null, PrettyPrinterAbstract $printer = null)
```
LukeTowers:
```suggestion
    public function __construct(array $ast, string $filePath = null, PrettyPrinterAbstract $printer = null)
```

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+    {
+        if (!($ast[0] instanceof Stmt\Return_)) {
+            throw new \InvalidArgumentException('configs must start with a return statement');
```
LukeTowers:
```suggestion
            throw new \InvalidArgumentException('ArrayFiles must start with a return statement');
```

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    /**
+     * Return a new instance of `ConfigFile` ready for modification of the file.
+     *
+     * @param string $file
+     * @param bool $createMissing
+     * @return ConfigFile|null
+     */
+    public static function read(string $file, bool $createMissing = false): ?ConfigFile
```
LukeTowers: Why would it be desired to interact with the ArrayFile parser without creating the file if it doesn't exist? I notice that the EnvFile parser always creates the file if it doesn't exist

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+    {
+        $exists = file_exists($file);
+
+        if (!$exists && !$createMissing) {
+            throw new \InvalidArgumentException('file not found');
+        }
+
+        $parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
```
LukeTowers: Does this cause issues with PHP 8 support?

LukeTowers: Side note, even if not, should we have a comment in here to switch it for the Laravel 9 upgrade?

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+
+        try {
+            $ast = $parser->parse(
+                $exists
+                    ? file_get_contents($file)
+                    : sprintf('<?php%1$s%1$sreturn [];%1$s', "\n")
```
LukeTowers: I know most of our config files look like
```php
<?php

return [
    // ...
];

```
any thoughts on instead making the default be more like
```php
<?php return [
    // ...
];

```

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+        return [($depth > 0) ? $pointer : null, $path];
+    }
+
+    /**
+     * Sort the config, supports: ConfigFile::SORT_ASC, ConfigFile::SORT_DESC, callable
+     *
+     * @param string|callable $mode
+     * @return ConfigFile
+     */
+    public function sort($mode = self::SORT_ASC): ConfigFile
```
LukeTowers: Do comments stick with the items when they're sorted / are comments retained by this parser?

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+        });
+    }
+
+    /**
+     * Write the current config to a file
+     *
+     * @param string|null $filePath
+     * @return void
+     */
+    public function write(string $filePath = null): void
```
LukeTowers: When the return type or parameter type is specified with PHP type hinting we should avoid duplicating it in the docblock

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    /**
+     * Returns a new instance of ConfigFunction
+     *
+     * @param string $name
+     * @param array $args
+     * @return ConfigFunction
+     */
+    public function function(string $name, array $args): ConfigFunction
```
LukeTowers: Should this be protected instead?

src/Config/ConfigFile.php
```diff
@@ -0,0 +1,428 @@
...
+        return new ConfigFunction($name, $args);
+    }
+
+    /**
+     * Returns a new instance of ConfigConst
+     *
+     * @param string $name
+     * @return ConfigConst
+     */
+    public function const(string $name): ConfigConst
```
LukeTowers: Should this be protected instead?

src/Config/ConfigFunction.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * @var string function name
+     */
+    protected $name;
+    /**
+     * @var array function arguments
+     */
+    protected $args;
```
LukeTowers:
```suggestion
    /**
     * @var string function name
     */
    protected $name;

    /**
     * @var array function arguments
     */
    protected $args;
```

src/Config/EnvFile.php
```diff
@@ -0,0 +1,268 @@
...
+        }
+    }
+
+    /**
+     * Parse a .env file, returns an array of the env file data and a key => pos map
+     *
+     * @param string $file
+     * @return array
+     */
+    protected function parse(string $file): array
```
LukeTowers: Can't we just use the DotEnv library to parse the file directly rather than doing it ourselves and potentially having a mismatch in behaviour?

LukeTowers: Or is the problem that doing it that way will eat the comments?

src/Config/EnvFile.php
```diff
@@ -0,0 +1,268 @@
...
+
+        return $env;
+    }
+
+    /**
+     * Get the default env file path
+     *
+     * @return string
+     */
+    public static function getEnvFilePath(): string
```
LukeTowers: Can this be protected?

src/Config/WinterPrinter.php
```diff
@@ -0,0 +1,51 @@
...
+
+        parent::__construct($options);
+    }
+
+    /**
+     * @param array $nodes
+     * @param bool $trailingComma
+     * @return string
+     */
+    protected function pMaybeMultiline(array $nodes, bool $trailingComma = false)
```
LukeTowers: Can we include the return type on the method signature instead of the docblock here?

src/Config/EnvFile.php
```diff
@@ -0,0 +1,268 @@
...
+        list($this->env, $this->map) = $this->parse($file);
+    }
+
+    /**
+     * Return a new instance of `EnvFile` ready for modification of the file.
+     *
+     * @param string|null $file
+     * @return EnvFile|null
+     */
+    public static function read(?string $file = null): ?EnvFile
```
LukeTowers:
```suggestion
    public static function read(?string $filePath = null): ?EnvFile
```

src/Config/EnvFile.php
```diff
@@ -0,0 +1,268 @@
...
+
+    /**
+     * @var array contains the env lookup map
+     */
+    protected $map = [];
+
+    /**
+     * @var string|null contains the filepath used to read / write
+     */
+    protected $file = null;
```
LukeTowers:
```suggestion
    protected $filePath = null;
```

LukeTowers (changes requested):

src/Config/ConfigFile.php
jaxwilko: In theory yes, but it would be worth writing a test to prove it.

src/Config/ConfigFile.php
jaxwilko: I hate it :joy: 

I'm of the opinion that the php file header should be on it's own line, it also looks nicer IMO

src/Config/ConfigFile.php
jaxwilko: It's there to allow for an exception to be thrown if a missing file is found, i.e. if you tried to edit `config/cms.php` and that file is missing, probably just throw an exception and be done.

By default we ship no env file, so it didn't really make sense to do it the same there

src/Config/ConfigFile.php
jaxwilko: PhpStorm complains about incomplete docblocks if you have a defined return type and no reference to it in the docblock, I can remove them if you want but I don't think it does much harm having them there

src/Config/ConfigFile.php
jaxwilko: Nope, it's meant for public use.

I.e.
```php
$config->set('winter.cms', $config->function('env', 'MY_ENV_KEY'));
```
renders to
```php
<?php

return [
    'winter' => [
        'cms' => env('MY_ENV_KEY')
    ],
];
```

src/Config/ConfigFile.php
jaxwilko: Same as above :)

src/Config/WinterPrinter.php
jaxwilko: Probably, I was just keeping it in sync with the parent `vendor/nikic/php-parser/lib/PhpParser/PrettyPrinter/Standard.php:1088`

image

src/Config/ConfigFile.php
LukeTowers: Fair enough 😜

src/Config/ConfigFile.php
LukeTowers: I've been removing them. Sounds like PHPStorm needs to figure its shit out, API docs generation shouldn't care which is used in order to identify the return type and my personal opinion is that duplicating them in the docblock is unnecessary and leads to maintenance headaches down the road when it gets changed in one place but not the other.

LukeTowers:
Blocked by comments on https://github.com/wintercms/winter/pull/404

LukeTowers (approved):

src/Config/ConfigWriter.php
```diff
@@ -2,217 +2,48 @@
...
-
-        // The target value to be replaced ($2)
-        $expression[] = '([^'.$quoteChar.']*)';
-
-        // The target key closure
-        $expression[] = '['.$quoteChar.']';
-
-        return '/' . implode('', $expression) . '/';
-    }
+        $parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7, $lexer = new Emulative([
```
bennothommo: @jaxwilko could you make sure the definition of `$lexer` happens on a separate line above? It makes it easier for debuggers to follow.

src/Foundation/Console/KeyGenerateCommand.php
```diff
@@ -3,80 +3,50 @@
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Foundation\Console\KeyGenerateCommand as KeyGenerateCommandBase;
```
bennothommo:
```suggestion
```

src/Foundation/Console/KeyGenerateCommand.php
```diff
@@ -3,80 +3,50 @@
...
-     * Set the application key in the config file.
+     * This method only asks for confirmation in production.
      *
-     * @param  string  $key
+     * @param  string  $warning
+     * @param  \Closure|bool|null  $callback
      * @return bool
      */
-    protected function setKeyInConfigFile($key)
+    public function confirmToProceed($warning = 'Application In Production!', $callback = null)
```
bennothommo, line 30: @jaxwilko does this confirmation implementation match the changes @LukeTowers made to command confirmations in the 1.2 branch recently?

src/Parse/Contracts/DataFileInterface.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Return a new instance of `DataFileInterface` ready for modification of the provided filepath.
+     */
+    public static function open(string $filePath): ?DataFileInterface;
+
+    /**
+     * Set a property within the data.
+     *
+     * @param string|array $key
+     * @param mixed|null $value
```
bennothommo: @jaxwilko Since we're targeting PHP 8, you can use union types for the parameters now ;)

src/Parse/Contracts/DataFileInterface.php
```diff
@@ -0,0 +1,29 @@
...
+     */
+    public function set($key, $value = null): DataFileInterface;
+
+    /**
+     * Write the current data to a file
+     *
+     * @param string|null $filePath
```
bennothommo: @jaxwilko If the parameter type hint indicates the necessary types, there's no need for the docblock to detail the parameter unless you want to add context after it.

src/Parse/Contracts/DataFileInterface.php
```diff
@@ -0,0 +1,29 @@
...
+     */
+    public function write(string $filePath = null): void;
```
bennothommo:
```suggestion
    public function write(?string $filePath = null): void;
```

src/Parse/EnvFile.php
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * @var array Map of variable names to line indexes
+     */
+    protected $map = [];
+
+    /**
+     * @var string|null Filepath currently being worked on
+     */
+    protected $filePath = null;
```
bennothommo: @jaxwilko Since we're targeting PHP 8, you could just put a type hint for these properties :)

src/Parse/EnvFile.php
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * EnvFile constructor
+     */
+    public function __construct(string $filePath)
+    {
+        $this->filePath = $filePath;
+
+        list($this->env, $this->map) = $this->parse($filePath);
```
bennothommo, line 33:
```suggestion
        [$this->env, $this->map] = $this->parse($filePath);
```

src/Parse/EnvFile.php
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * @var string|null Filepath currently being worked on
+     */
+    protected $filePath = null;
+
+    /**
+     * EnvFile constructor
+     */
+    public function __construct(string $filePath)
```
bennothommo: @jaxwilko discovered this using PHPStan, but if you don't make the constructor API concrete, it can make creating static instances a problem if people extend the class and change up the API. Simple fix is to make the constructor final if you don't intend for people to change the constructor, even after extending.

```suggestion
    final public function __construct(string $filePath)
```

LukeTowers:
Still some issues @jaxwilko:

```php
<?php

return [

    /*
    |--------------------------------------------------------------------------
    | Default Broadcaster
    |--------------------------------------------------------------------------
    |
    | This option controls the default broadcaster that will be used by the
    | framework when an event needs to be broadcast. You may set this to
    | any of the connections defined in the "connections" array below.
    |
    | Supported: "pusher", "ably", "redis", "log", "null"
    |
    */

    'default' => env('BROADCAST_DRIVER', 'null'),

    /*
    |--------------------------------------------------------------------------
    | Broadcast Connections
    |--------------------------------------------------------------------------
    |
    | Here you may define all of the broadcast connections that will be used
    | to broadcast events to other systems or over websockets. Samples of
    | each available type of connection are provided inside this array.
    |
    */

    'connections' => [

        'pusher' => [
            'app_id'         => env('PUSHER_APP_ID'),
            'client_options' => [
                // Guzzle client options: https://docs.guzzlephp.org/en/stable/request-options.html
            ],
            'driver'         => 'pusher',
            'key'            => env('PUSHER_APP_KEY'),
            'options'        => [
                'cluster' => env('PUSHER_APP_CLUSTER'),
                'useTLS'  => true,
            ],
            'secret'         => env('PUSHER_APP_SECRET'),
        ],

        'ably' => [
            'driver' => 'ably',
            'key'    => env('ABLY_KEY'),
        ],

        'redis' => [
            'connection' => 'default',
            'driver'     => 'redis',
        ],

        'log' => [
            'driver' => 'log',
        ],

        'null' => [
            'driver' => 'null',
        ],

    ],

];
```

Turned into 

```php
<?php

return [

    /*
    |--------------------------------------------------------------------------
    | Default Broadcaster
    |--------------------------------------------------------------------------
    |
    | This option controls the default broadcaster that will be used by the
    | framework when an event needs to be broadcast. You may set this to
    | any of the connections defined in the "connections" array below.
    |
    | Supported: "pusher", "ably", "redis", "log", "null"
    |
    */

    'default' => env('BROADCAST_DRIVER', 'null'),

    /*
    |--------------------------------------------------------------------------
    | Broadcast Connections
    |--------------------------------------------------------------------------
    |
    | Here you may define all of the broadcast connections that will be used
    | to broadcast events to other systems or over websockets. Samples of
    | each available type of connection are provided inside this array.
    |
    */

    'connections' => [
        'pusher' => [
            'app_id' => env('PUSHER_APP_ID'),
            'client_options' => [
                // Guzzle client options: https://docs.guzzlephp.org/en/stable/request-options.html
            ],
            'driver' => 'pusher',
            'key' => env('PUSHER_APP_KEY'),
            'options' => [
                'cluster' => env('PUSHER_APP_CLUSTER'),
                'useTLS' => true,
            ],
            'secret' => env('PUSHER_APP_SECRET'),
            // Guzzle client options: https://docs.guzzlephp.org/en/stable/request-options.html
        ],
        'ably' => [
            'driver' => 'ably',
            'key' => env('ABLY_KEY'),
        ],
        'redis' => [
            'connection' => 'default',
            'driver' => 'redis',
        ],
        'log' => [
            'driver' => 'log',
        ],
        'null' => [
            'driver' => 'null',
        ],
        // Guzzle client options: https://docs.guzzlephp.org/en/stable/request-options.html
    ],
];
```
Note all of the extra injections of the `// Guzzle client options` comment. Same issue with the single line comments on `cache.php`.

src/Parse/EnvFile.php
```diff
@@ -0,0 +1,238 @@
...
+     * // or
+     * $env->set([
+     *     'APP_PROPERTY' => 'example',
+     *     'DIF_PROPERTY' => 'example'
+     * ]);
+     * ```
+     * @param array|string $key
+     * @param mixed|null $value
+     */
+    public function set($key, $value = null): EnvFile
```
bennothommo:
```suggestion
    public function set(array|string $key, $value = null): EnvFile
```

src/Parse/EnvFile.php
LukeTowers: I personally prefer using list() rather than just square brackets, easier to google what the heck's happening for new devs.

bennothommo: Fair point.

src/Foundation/Console/KeyGenerateCommand.php
LukeTowers: This command extends the Laravel base command, not Winter's. This is only implemented here to override the default message shown to the user when the command is run and a key exists. The confirmation behaviour I recently added is available through a trait, but we don't require confirmation with input in this case, just yes / no.

PR #42, opened 2021-07-12
Add support for trusted proxies

### LINKED ISSUE #232: Add support for trusted proxies
Laravel 6 has support for trusted proxies (https://laravel.com/docs/6.x/requests#configuring-trusted-proxies) which allows certain setups (such as AWS Elastic Load Balancing, Heroku, Gitpod) where TLS/SSL communications are terminated at the entrypoint to still use HTTPS links even though the actual PHP is not output in a TLS/SSL request.

We currently do not have a similar capacity in Winter CMS, and should probably implement it in similar fashion to the trusted hosts functionality. While it is certainly possible to implement this through a plugin, it might be better being a first-class feature in order for Winter CMS to just "work" in more complex environments.

For example, at the moment, I am unable to access the administration panel of a fresh Winter CMS install in a Gitpod instance, even after setting all possible configuration options for the app URL (`app.url`, `APP_URL`, etc.), because the Symfony `Request` class interprets my request as a non-secure request and converts the protocol of all my redirects, links and assets to `http`. Gitpod, however, communicates through a `https` URL, therefore, at the very least, I have an unstyled login screen - at worse, a redirect loop.

When adding a custom middleware through a plugin to use trusted proxies, which are then populated in the request in a similar fashion to the Laravel docs link above, it works.

Laravel uses the `fideloper/proxy` library to provide this functionality - we can either do the same, or roll our own.

bennothommo:
Fixes https://github.com/wintercms/winter/issues/232.

See the issue for more info.

### DIFF
--- a/src/Foundation/Http/Kernel.php
+++ b/src/Foundation/Http/Kernel.php
@@ -28,6 +28,7 @@ class Kernel extends HttpKernel
      */
     protected $middleware = [
         \Winter\Storm\Foundation\Http\Middleware\CheckForTrustedHost::class,
+        \Winter\Storm\Foundation\Http\Middleware\CheckForTrustedProxies::class,
         \Winter\Storm\Foundation\Http\Middleware\CheckForMaintenanceMode::class,
     ];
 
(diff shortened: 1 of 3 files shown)


LukeTowers:
LGTM

PR #44, opened 2021-07-21
Retrieve and store headers if Http::toFile() method is used.

### LINKED ISSUE #31: Add support for specifying headers on Http::toFile()
Refs: octobercms/october#5453

bennothommo:
Previously, the `Http::toFile()` method would prevent headers from being read, and would simply be discarded. This adds a header stream and writes the headers to the corresponding property even if `toFile` is used, keeping parity with a non-file request.

This change also adds trimming to the headers.

Fixes https://github.com/wintercms/winter/issues/31

### DIFF
--- a/src/Network/Http.php
+++ b/src/Network/Http.php
@@ -253,7 +253,10 @@ class Http
 
     /**
      * Execute the HTTP request.
-     * @return string response body
+     *
+     * Returns the Http instance in order to be able to inspect and retrieve the response.
+     *
+     * @return self
      */
     public function send()
     {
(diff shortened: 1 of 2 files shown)


src/Network/Http.php
```diff
@@ -347,6 +357,9 @@ public function send()
         curl_close($curl);
 
         if ($this->streamFile) {
+            rewind($headerStream);
```
LukeTowers, line 360: Are there any concerns here with the temp stream being modified in between `curl_exec` being called and this code being run? Should we namespace the stream at all? Is that even possible?

LukeTowers:
Looks fantastic @bennothommo, much better than the other "fix" for the original issue. Just have the one question about it, otherwise it should be good to go.

src/Network/Http.php
bennothommo: @LukeTowers according to the [PHP docs](https://www.php.net/manual/en/wrappers.php.php), temp streams are not reusable, so it would appear to me that each stream operates on its own. There's no namespaces for it either, which if the previous point is true would make them redundant anyway.

LukeTowers: Tested it with the following and it seems to be accurate so sounds good to me:

```php
$stream1 = fopen('php://temp', 'r+');
$stream2 = fopen('php://temp', 'r+');

fwrite($stream1, 'test');

rewind($stream2);
stream_get_contents($stream2);
```

PR #45, opened 2021-07-27
Improvements to URL generation

LukeTowers:
By default Laravel's URL generation does not include any validation or escaping that ensures URLs generated by it are valid URLs, it merely takes the role of a simple string manipulator.

After considerable research and development this change will flesh out the URL generation functionality in Winter CMS to ensure that return values from the URL generator are valid URLs and are properly escaped and safe to render in HTML.

This covers several cases including increased compliance with `http_build_url()`, always including a slash after the hostname, removing dot segments as required, and support for properly encoding all parts of the URL.

References:
- https://datatracker.ietf.org/doc/html/rfc3986/
- https://www.anchor.com.au/blog/2013/06/safely-escaping-and-encoding-urls/
- https://github.com/m6w6/ext-http/blob/master/src/php_http_url.c
- https://php.uz/manual/en/function.http-build-url.php
- https://github.com/jakeasmith/http_build_url
- https://github.com/ivantcholakov/http_build_url


src/Router/UrlGenerator.php
```diff
@@ -47,40 +62,120 @@ public static function buildUrl($url, $replace = [], $flags = HTTP_URL_REPLACE,
...
+                    $value = (int) $value;
+                    if ($value < 0 || $value > 65535) {
+                        unset($url['key']);
+                    }
+                }
+            }
+        };
+        if (!is_iterable($url) || !is_iterable($replace)) {
+            // dd($url, $replace, $flags);
+        }
```
LukeTowers: need to remove

src/Router/UrlGenerator.php
```diff
@@ -1,39 +1,54 @@
...
-        }
-        if (is_string($replace)) {
-            $replace = parse_url($replace);
-        }
-
         $urlSegments = ['scheme', 'host', 'user', 'pass', 'port', 'path', 'query', 'fragment'];
 
+        // Setup special handling for common schemes
+        // @see https://en.wikipedia.org/wiki/List_of_URI_schemes for future special cases
+        $singleColonSchemes = ['mailto', 'tel', 'sms'];
```
bennothommo, line 51: I think we might need to try and cater for *any* type of single-colon scheme, as opposed to being explicit. The only one we'd really need to do some sort of handling (or more to the point, denying) is `javascript:`.

src/Router/UrlGenerator.php
```diff
@@ -47,40 +62,120 @@ public static function buildUrl($url, $replace = [], $flags = HTTP_URL_REPLACE,
...
+        if (is_string($url)) {
+            $url = parse_url($url);
+        }
+        if (is_string($replace)) {
+            $replace = parse_url($replace);
+        }
+
+        // Prepare input data
+        $cleanUrlArray = function (&$url) use ($urlSegments) {
+            if (!is_array($url)) {
```
bennothommo, line 76: I don't think this check would be needed. At the point that this function is used, both `$url` and `$replace` would have to be arrays.

src/Router/UrlGenerator.php
```diff
@@ -47,40 +62,120 @@ public static function buildUrl($url, $replace = [], $flags = HTTP_URL_REPLACE,
...
+                $url = [];
+            }
+
+            foreach ($url as $key => &$value) {
+                // Remove invalid segments
```
bennothommo, line 81: As far as I can tell, `$urlSegments` is not being used as a parameter and thus this segment removal is probably redundant. Segment removal seems to be happening on line 180 onwards.

src/Router/UrlGenerator.php
```diff
@@ -47,40 +62,120 @@ public static function buildUrl($url, $replace = [], $flags = HTTP_URL_REPLACE,
...
-                $urlQuery = $joinedQuery = [];
+            if (
+                ($flags & HTTP_URL_JOIN_QUERY) &&
+                isset($url['query']) &&
+                isset($replace['query'])
+            ) {
+                $uQuery = $url['query'];
+                $rQuery = $replace['query'];
+
+                if (!is_array($uQuery)) { parse_str($uQuery, $uQuery); }
```
bennothommo:
```suggestion
                if (!is_array($uQuery)) {
                    parse_str($uQuery, $uQuery);
                }
```

src/Router/UrlGenerator.php
```diff
@@ -47,40 +62,120 @@ public static function buildUrl($url, $replace = [], $flags = HTTP_URL_REPLACE,
...
+                if (!is_array($rQuery)) { parse_str($rQuery, $rQuery); }
```
bennothommo:
```suggestion
                if (!is_array($rQuery)) {
                    parse_str($rQuery, $rQuery);
                }
```

bennothommo:
@LukeTowers it appears the overwrite of the `to()` method in `UrlGenerator` is having a knock-on effect in that all URLs being generated now have a trailing forward-slash if no path is specified for the URL.

Where previously the `to()` method would return `https://wintercms.com`, it now returns `https://wintercms.com/`.

Should we allow for URLs to be created without trailing slashes?

src/Router/UrlGenerator.php
LukeTowers: The reason I'm explicit here is because Laravel is explicit, probably because these are uncommon but still relatively common schemes to use in a web environment. I'm not sure what a generic approach to this would look like without getting really into the weeds, at some point we have to accept that it's impossible to handle every conceivable variation of custom schemes that could be out there and just focus on the ones most likely to affect our users.

src/Router/UrlGenerator.php
LukeTowers: Nope, if something that isn't a string or an array is passed, then it's not an array. I.e. false or null would be the main offenders.

src/Router/UrlGenerator.php
LukeTowers: It's not a parameter, just a way to ensure only valid URL segments are in the $url and $replace properties. Line 180 has to do with stripping out valid but undesired segments.

src/Router/UrlGenerator.php
bennothommo: @LukeTowers if `false` or `null` are accepted, then the docblock will need to be updated, as it only says `array|string` at the moment.

src/Router/UrlGenerator.php
bennothommo: Gotcha

LukeTowers:
@bennothommo looking at a few resources online it would appear that it doesn't really have a negative effect with or without the trailing slash after just the hostname (see https://ahrefs.com/blog/trailing-slash/).

As far as Laravel automatically trimming the trailing slash on all URLs, looks like they've been doing it since the very first commit present on GitHub, although there was a commit in 5.0.30 that fixed it so that `Url::to('/')` would output `http://example.com/` instead of `http://example.com` (https://github.com/laravel/framework/commit/9b82fae3feaedb004971584166aa0468d9cd86e4) and then in typical Taylor fashion an unexplained commit that reverted that behavior (https://github.com/laravel/framework/commit/cf24df89336316e74b50d2973b040657cdfafef5#diff-1c99b8c24a0f2a0d87cf056dffb3a4504825a2f7fb4fe55dbdfb5d86742f3c5aR93)

Personally I'd prefer if URL generation enforced slash after hostname always simply because if there's no path the hostname and the query and/or the fragment look strange smushed together (i.e. `https://example.com?query` or `https://example.com#query`). We can probably update the other tests that are now failing as a result of the change and include a note in the API changes section.

bennothommo:
@LukeTowers my only "devil's advocate" argument would be someone using the `to()` method as a prefix, and adding their own path afterwards outside of the method.

No idea why people would do that - so yeah, I got nothing. Let's do it your way ;)

bennothommo:
Tests are now fixed.

src/Router/UrlGenerator.php
LukeTowers: Done

LukeTowers:
@bennothommo if you have some time, could you take a stab at adding tests for https://github.com/wintercms/storm/pull/45/files#diff-53e0edf8e6b8ab3280f8b8dfa913447667e117fd74dc1c5849412f8793045b12R243-R248

PR #47, opened 2021-08-21
Impersonation improvements

LukeTowers:
This add several improvements to the impersonation functionality present in Winter CMS:

- The `User` model now has a new method: `canBeImpersonated()` that receives the `$impersonator` if there is one and `false` if there isn't. This allows the impersonated user to decide if they want to be impersonated or not and makes it easy to extend the default logic for whatever custom system is desired.
- The `model.user.beforeImpersonate` event has been changed to a halting event so that third party plugins are able to override the default return values from `canBeImpersonated()` to implement more or less strict impersonation protection policies as desired on a per project basis by returning a boolean flag indicating if the user can be impersonated or not
- The `getMergedPermissions()` method in `Backend\Models\User` now will filter available permissions by the current impersonator to prevent impersonators from having more access to the system than they otherwise would have.

>**NOTE**: Feedback would be highly appreciated on this and the related PRs in wintercms/winter and wintercms/wn-user-plugin.

Related PRs:
- wintercms/winter#273
- wintercms/wn-user-plugin#10

Below is a brain dump on the subject along with some references of how other systems implement user impersonation:

# Auth Manager notes:

## User Impersonation:

Use Cases:
- Testing and debugging without requiring access to other user's credentials or having to go through the login flow

- Testing access levels
- Seeing what the user sees for support
- Allowing system services to act as a user

Current Implementation:
- Logic for all implementations of AuthManager handled in base AuthManager
- Backend:
    - Session Key: admin_auth
    - Model: Backend\Models\User
    - Users controller: update_onImpersonateUser($id);
        - requires `backend.impersonate_users` permission as only layer of protection
        - redirects to backend user myaccount page when done
- Winter.User:
    - Session Key: user_auth
    - Model: Winter\User\Models\User
    - Implemented in https://github.com/wintercms/wn-user-plugin/commit/048b155329ef39623708be468324da2a88495922
    - Meant solely for backend users to "impersonate" (i.e. just login as frontend users)
    - Impersonate method called when a user might not even be logged in under the user_auth manager
    - Requires `winter.users.impersonate_user` permission

Considerations:
- Always need to be able to identify an impersonated user as being impersonated
- Only authorized users should be able to impersonate other users
- Logs must be able to correctly identify the actual user operating the account
- Multiple authentication systems can be using the same code at the same time
- Read only access for impersonating users may be desired in some cases but undesired in others
- Users shouldn't be able to impersonate themselves
- Users shouldn't be able to impersonate someone with more access than they have
- It should be obvious when you are impersonating another person's account
    - Header banner
    - Display of username includes (Impersonating) text
- how to handle impersonating of users that shouldn't have access (deleted / archived, unactivated, banned, throttled, etc)

Other examples in the wild:
- su command in unix
- https://github.com/viacreative/sudo-su
    - stores flag when impersonating
    - stores original user ID in session
    - provides way to get the original user and identify when impersonating
    - configuration for what the user model is, assumes imperonsator and impersonatee are the same model
    - user input = impersonateeID & impersonatorID
- https://github.com/OctopyID/LaraPersonate
    - stores previous user
    - checks if user can be impersonated and if impersonater can impersonate
    - configuration for what the user model is, assumes imperonsator and impersonatee are the same model
- https://github.com/404labfr/laravel-impersonate
    - uses a trait on the user model
    - provides route middleware for preventing access to routes when the user is being impersonated
- https://docs.laravel-enso.com/backend/impersonate.html#installation
- https://pineco.de/impersonating-users/
    - "Sometimes we face situations when all the tests are passing, we find no bugs, but still, on our users’ part, something broke. By impersonating our users, we can see what they see and track the bugs down easily."
- https://mauricius.dev/easily-impersonate-any-user-in-a-laravel-application/
    - "see the application from their point of view, without having to log out and log in again"
    - "you want to recreate a bug encountered by one of your users, without having them to share their password with you."
    - Auth::once() logs in without session or cookies
- https://tenancyforlaravel.com/docs/v3/features/user-impersonation/
- https://help.channeltivity.com/support/solutions/articles/3000066767-what-is-user-impersonation-and-how-do-i-impersonate-a-user-
    - Role based
- https://cloud.google.com/iam/docs/impersonating-service-accounts
    - Explicit configuration per user of what user roles / records they are permitted to act as
    - List of members that can impersonate the account on the account itself
    - https://cloud.google.com/iam/docs/impersonating-service-accounts#allow-impersonation
- https://docs.servicenow.com/bundle/quebec-platform-administration/page/administer/users-and-groups/concept/c_ImpersonateAUser.html
    - Logs actions taken as being done by the impersonatee
    - Permissions the impersonator doesn't have access to are not avaiable while acting as the impersonatee (even if the impersonatee has those permissions)
    - impersonators can't update the password (or presumably the email address of impersonatees)
- https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation
- https://docs.microsoft.com/en-us/exchange/client-developer/exchange-web-services/impersonation-and-ews-in-exchange
- https://opendistro.github.io/for-elasticsearch-docs/docs/security/access-control/impersonation/
- https://support.google.com/admanager/answer/1241070?hl=en ("Impersonator as Impersonatee" in account view)
- https://help.docebo.com/hc/en-us/articles/360020125239-Impersonating-a-User-or-a-Power-User

### DIFF
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -736,7 +736,7 @@ class Manager implements \Illuminate\Contracts\Auth\StatefulGuard
 
         if (!$canImpersonate) {
             // @TODO: translate / make exception more specific
-            throw new Exception("You cannot impersonate the selected user.");
+            throw new Exception('You cannot impersonate the selected user.');
         }
 
         // Impersonate the requested user by becoming them in the request & the session
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -705,10 +705,16 @@ class Manager implements \Illuminate\Contracts\Auth\StatefulGuard
      */
     public function impersonate($impersonatee)
     {
-        // Get the current user, i.e. the "impersonator"
-        $userArray = $this->getPersistCodeFromSession();
-        $impersonatorId = $userArray ? $userArray[0] : null;
-        $impersonator = $impersonatorId ? $this->findUserById($impersonatorId) : false;
+        // If the session is already being impersonated, then use the original impersonator
+        if ($this->isImpersonator()) {
+            $impersonator = $this->getImpersonator() ?: false;
+            $impersonatorId = $impersonator ? $impersonator->id : null;
+        } else {
+            // Get the current user
+            $userArray = $this->getPersistCodeFromSession();
+            $impersonatorId = $userArray ? $userArray[0] : null;
+            $impersonator = $impersonatorId ? $this->findUserById($impersonatorId) : false;
+        }
 
         /**
          * @event model.auth.beforeImpersonate
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -828,4 +828,18 @@ class Manager implements \Illuminate\Contracts\Auth\StatefulGuard
 
         return $this->createUserModel()->find($impersonatorId);
     }
+
+    /**
+     * Gets the user for the request, taking into account impersonation
+     *
+     * @return mixed (Models\User || null)
+     */
+    public function getRealUser()
+    {
+        if ($impersonator = $this->getImpersonator()) {
+            return $impersonator;
+        } else {
+            return $this->getUser();
+        }
+    }
 }
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -700,6 +700,7 @@ class Manager implements \Illuminate\Contracts\Auth\StatefulGuard
      * Impersonates the given user and sets properties in the session but not the cookie.
      *
      * @param Models\User $impersonatee
+     * @throws Exception If the current user is not permitted to impersonate the provided user
      * @return void
      */
     public function impersonate($impersonatee)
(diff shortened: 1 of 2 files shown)


src/Auth/Manager.php
```diff
@@ -712,16 +713,31 @@ public function impersonate($impersonatee)
...
          */
-        $impersonatee->fireEvent('model.auth.beforeImpersonate', [$impersonator]);
+        $canImpersonate = $impersonatee->fireEvent('model.auth.beforeImpersonate', [$impersonator], true);
+        if (is_null($canImpersonate)) {
+            $canImpersonate = $impersonatee->canBeImpersonated($impersonator);
+        }
+
+        if (!$canImpersonate) {
+            // @TODO: translate / make exception more specific
+            throw new Exception("You cannot impersonate the selected user.");
```
bennothommo:
```suggestion
            throw new Exception('You cannot impersonate the selected user.');
```

bennothommo:
@LukeTowers I've had a look at this more thoroughly on the whole context of changes. I think my biggest worry with this is we're putting a strict opinion on impersonation.

A couple of thoughts:
- While the permission restrictions make sense from a security standpoint (the impersonator cannot get permissions from an impersonatee that they don't have themselves), it would prevent certain use cases from working. For example, you mention technical support as a use-case, but say there's a problem with the blog and the support staff member doesn't have access to the blog - if they then impersonate an author, they have no way of getting to the blog to assist. I think we can't make this decision for everyone - there needs to be a way to control whether the permissions get restricted or not.
- While the original intention of the impersonation in terms of the User plugin was that only a Backend user could impersonate a front-end User, I don't think there was anything technically stopping people from implementing impersonation in the front-end, say as a plugin, if they so choose. With your PR to the User plugin, this scenario would now be enforced, so this technically is a (minor) breaking change.
- Further to my comments on the PR to the main repo - should we perhaps establish a method that determines where the user is returned when impersonation ceases? If the goal is to allow inter-op between auth streams and impersonation, this might be useful.

PR #48, opened 2021-08-31
create onSuspend abstract event

josephcrowell:
can be used within winter.user plugin to fire user suspension notifications


src/Auth/Models/Throttle.php
```diff
@@ -81,6 +81,11 @@ public function addLoginAttempt()
 
         if ($this->getLoginAttempts() >= static::$attemptLimit) {
             $this->suspend();
+
+            /*
+            * Fire the user 'onSuspend' event
+            */
```
bennothommo:
```suggestion
             * Fire the user 'onSuspend' event
             */
```

src/Auth/Models/Throttle.php
```diff
@@ -267,4 +272,4 @@ public function getIsBannedAttribute($banned)
     {
         return (bool) $banned;
     }
-}
+}
```
bennothommo: PHP files must terminate with a blank line as per our code style guidelines.

```suggestion
}

```

src/Auth/Models/User.php
```diff
@@ -148,6 +148,10 @@ public function afterLogin()
         $this->forceSave();
     }
 
+    public function onSuspend()
```
bennothommo, line 154: @josephcrowell Please add a docblock here that describes the intention of the method.

bennothommo (changes requested):

src/Auth/Models/Throttle.php
```diff
@@ -81,6 +81,11 @@ public function addLoginAttempt()
 
         if ($this->getLoginAttempts() >= static::$attemptLimit) {
             $this->suspend();
+
+            /*
+             * Fire the user 'onSuspend' event
+             */
+            $this->user()->onSuspend();
```
LukeTowers, line 88: This would have to be `$this->getUser()` or `$this->user`, `$this->user()` refers to a BelongsTo relationship instance which passes method calls off to a QueryBuilder.

PR #53, opened 2021-11-03
Allow for YAML rendering pre/post processing.

bennothommo:
This is a BC break since it changes the contract. Classes implementing the `Winter\Storm\Parse\Processor\Contracts\YamlProcessor` interface should instead extend the `Winter\Storm\Parse\Processor\YamlProcessor` abstract to maintain parity with the interface.

This change is in preparation of a change that I intend to make for the Builder plugin to ensure that the `version.yaml` uses quoted version keys when generating the migrations.


LukeTowers:
How does this interact with #21 and the related PR?

bennothommo:
It should be entirely compatible with that PR (besides changing the processor to extend the abstract class as opposed to implementing the interface). These new methods are for rendering though, so while that PR ensures that our older version files can still be read, this will allow us to ensure that the YAML files are correctly written as well.

LukeTowers:
@bennothommo what's the status of this?

bennothommo:
@LukeTowers it should be ready to go.

LukeTowers:
@bennothommo looks like the tests need to be fixed for this PR

bennothommo:
@LukeTowers fixed up the tests

PR #54, opened 2021-11-17
Add create:class command

rezalaal:


LukeTowers:
I'm not sure a scaffolding command is needed for generating a single file that wouldn't have any boilerplate, could you expand on your reasons for wanting this included?

rezalaal:
There are many times when we create a class and encounter the error that the class was not found. Because it is sensitive to letters or for other reasons. In this way, both the construction of the class is available and typing errors are prevented when creating the class, and this command can be extended in the future.
Thank you

LukeTowers:
The class not found errors, do you get them when you name the directory Classes instead of classes? If so we can probably instead tweak the autoloading code to allow for both forms since I know titled case folders are common in other projects.

rezalaal:
Yes I get it when class name and class folder  are same.

LukeTowers:
Can you provide a specific example of a class that fails to load in Winter with the exact path to the class and the contents of the class file? This sounds like a separate issue that adding a scaffolding command doesn't really fix.

rezalaal:
Now that I want to send a sample, everything works fine ;-)

LukeTowers:
@rezalaal of course 😉 I'll close this for now but feel free to submit a new issue if you encounter issues in the future!

PR #55, opened 2021-11-24
Extend withDeferred() to allow pivot conditions

lex0r:
If a pivot has additional fields that need to be considered when fetching related data using withDeferred() there must be a way to add those conditions. The original relation query is modified by withDeferred() in such a way that it's impossible to keep any where() clauses added to it. Hence this workaround, which may not seem elegant but does the job.


LukeTowers:
Could you share your custom query that was failing with the current state of withDeferred() so we can see if we can come up with a more robust solution?

LukeTowers:
Closing due to lack of response from @lex0r

PR #59, opened 2021-12-06
Updated DotEnv call to pass repository

jaxwilko:
Adds a patch to update DotEnv usage to be compatible with changes introduced in this PR https://github.com/vlucas/phpdotenv/pull/387

I'm not sure if this is how we actually want to handle env repository long term, but in the short term this'll get the application booting :)


LukeTowers:
Should we make our dependency on dotenv more explicit in composer.json so that it's more obvious what version of dotenv we're using, even if it's just matching Laravel's own statement?

Also is there a reference to any changes that Laravel made to support this too?

PR #60, opened 2021-12-06
Updated artisan service provider to be compatible with laravel

jaxwilko:
In newer versions of laravel, the default application commands are lazy loaded in. Because of this they've switched from using the facade accessor to the class string. This change was introduced in this commit: https://github.com/laravel/framework/commit/82b212c1d797e73f658ef461b11796ca75707361

I may have missed some strings that need updating as I'm just running unit tests and fixing bugs as they pop up, but this should be a start :)

### DIFF
--- a/src/Foundation/Providers/ArtisanServiceProvider.php
+++ b/src/Foundation/Providers/ArtisanServiceProvider.php
@@ -19,7 +19,7 @@ class ArtisanServiceProvider extends ArtisanServiceProviderBase
         'ConfigClear'           => \Illuminate\Foundation\Console\ConfigClearCommand::class,
         'Down'                  => \Illuminate\Foundation\Console\DownCommand::class,
         'Environment'           => \Illuminate\Foundation\Console\EnvironmentCommand::class,
-        'KeyGenerate'           => KeyGenerateCommand::class,
+        'KeyGenerate'           => \Winter\Storm\Foundation\Console\KeyGenerateCommand::class,
         'Optimize'              => \Illuminate\Foundation\Console\OptimizeCommand::class,
         'PackageDiscover'       => \Illuminate\Foundation\Console\PackageDiscoverCommand::class,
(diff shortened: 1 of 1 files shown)


src/Foundation/Providers/ArtisanServiceProvider.php
```diff
@@ -12,32 +12,36 @@ class ArtisanServiceProvider extends ArtisanServiceProviderBase
...
-        'Up'              => 'command.up',
-        'ViewClear'       => 'command.view.clear',
+        'CacheClear'            => \Illuminate\Cache\Console\ClearCommand::class,
+        'CacheForget'           => \Illuminate\Cache\Console\ForgetCommand::class,
+        'ClearCompiled'         => \Winter\Storm\Foundation\Console\ClearCompiledCommand::class,
+        'ConfigCache'           => \Illuminate\Foundation\Console\ConfigCacheCommand::class,
+        'ConfigClear'           => \Illuminate\Foundation\Console\ConfigClearCommand::class,
+        'Down'                  => \Illuminate\Foundation\Console\DownCommand::class,
+        'Environment'           => \Illuminate\Foundation\Console\EnvironmentCommand::class,
+        'KeyGenerate'           => KeyGenerateCommand::class,
```
LukeTowers: Use the full namespace here and remove the use statements from the top of the file please.

jaxwilko: Can do, only reason I didn't was that the class was already being included for use later on in the file, I can swap both usages to a full path tho :)

LukeTowers: Ah, gotcha. If that's the case then you can leave the import if you want but just make this reference a full path. Up to you, just want full paths here so it's more obvious which are being overridden by Winter.

jaxwilko: Added full path in b58be85

PR #61, opened 2021-12-08
Add SVG utility

bennothommo:
This PR adds an SVG utility that can be used to extract and sanitize an SVG from a given path.

This class is a precursor to allowing full support for SVGs in Winter.


LukeTowers:
@bennothommo would this make more sense under the Winter\Storm\Parse namespace?

bennothommo:
@LukeTowers it's not really a parser though. It's more a utility, like our `Str` or `Arr` utilities.

composer.json
```diff
@@ -44,7 +44,8 @@
         "league/csv": "~9.1",
         "nesbot/carbon": "^2.0",
         "laravel/framework": "~6.0",
-        "laravel/tinker": "~2.0"
+        "laravel/tinker": "~2.0",
+        "enshrined/svg-sanitize": "^0.14.1"
```
LukeTowers: Why so restrictive on the version? I think they're on 0.15 now

src/Support/Svg.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    /**
+     * Extracts and sanitizes SVG code from a given file.
+     *
+     * @param string $path The path to the SVG file.
+     * @param boolean $minify Whether to minify the extracted SVG code.
+     * @return string
+     * @throws ApplicationException If no file, or a malformed SVG, is found at the given path.
+     */
+    public static function extract($path, $minify = true)
```
LukeTowers, line 24: Why Svg::extract() and not Svg::sanitize()?

bennothommo: The intention of the utility is to provide the (sanitized) SVG markup for use, either directly in template code or to be populated into a file, hence why I went with `extract`.

composer.json
bennothommo: I think that was inserted by `composer require`, I'll fix it up.

LukeTowers:
@bennothommo can you add a PR to the docs for this?

bennothommo:
@LukeTowers https://github.com/wintercms/docs/pull/90

PR #63, opened 2021-12-14
Add default name to commands for lazy loading

bennothommo:
Refs:
- https://github.com/laravel/framework/pull/34873
- https://github.com/wintercms/winter/pull/148

### DIFF
--- a/src/Scaffold/Console/CreateTheme.php
+++ b/src/Scaffold/Console/CreateTheme.php
@@ -8,11 +8,11 @@ use Symfony\Component\Console\Input\InputArgument;
 class CreateTheme extends GeneratorCommand
 {
     /**
-     * The console command name.
+     * The default command name for lazy loading.
      *
-     * @var string
+     * @var string|null
      */
-    protected $name = 'create:theme';
+    protected static $defaultName = 'create:theme';
 
     /**
      * The console command description.
--- a/src/Scaffold/Console/CreateTheme.php
+++ b/src/Scaffold/Console/CreateTheme.php
@@ -2,8 +2,6 @@
 
 use Exception;
 use Winter\Storm\Scaffold\GeneratorCommand;
-use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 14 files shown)


LukeTowers:
What needs to be done in order to support this for plugin registered commands?

bennothommo:
@LukeTowers I think if they define `$defaultName` as a property with the commands, the lazy loading will kick in automatically. At the very least, I don't believe they've disabled eager-loading commands in the old fashion.

src/Scaffold/Console/command/command.stub
```diff
@@ -9,7 +9,7 @@ class {{studly_name}} extends Command
     /**
      * @var string The console command name.
      */
-    protected $name = '{{lower_plugin}}:{{lower_name}}';
+    protected static $defaultName = '{{lower_plugin}}:{{lower_name}}';
```
LukeTowers, line 12: I think we should probably include the signature property like Laravel is doing in their scaffolding

src/Scaffold/Console/CreateTheme.php
```diff
@@ -8,11 +8,11 @@
...
 {
     /**
-     * The console command name.
+     * The default command name for lazy loading.
      *
-     * @var string
+     * @var string|null
      */
-    protected $name = 'create:theme';
+    protected static $defaultName = 'create:theme';
```
LukeTowers, line 13: What was the name property being used by previously and are we save to no longer use it?

src/Scaffold/Console/command/command.stub
bennothommo: Sounds good.

src/Scaffold/Console/CreateTheme.php
bennothommo: I believe `$name` is used in the absence of a `$signature`. It's safe to remove if we use `$defaultName` or `$signature`, as far as I can tell.

PR #64, opened 2021-12-14
Code analysis for Laravel 9 branch

bennothommo:
Thought I would try installing Larastan/PHPStan and run some code analysis to look for some invalid PHP docblocks and dead code and the like. So far, it's picked up about 1100 errors - not sure if it's a bridge too far to cross, but interesting nonetheless.

We're testing against Level 5 so far, which seems to be a good compromise between ensuring all docblocks and signatures are correct whilst not having PHPStan being pedantic about *everything*.

## Modules

- [x] Argon
- [x] Auth
- [x] Config
- [x] Cookie
- [x] Database
- [x] Events
- [x] Exception
- [x] Extension
- [x] Filesystem
- [x] Flash
- [x] Foundation
- [x] Halcyon
- [x] Html
- [x] Http
- [x] Mail
- [x] Network
- [x] Parse
- [x] Redis
- [x] Router
- [x] Scaffold
- [x] Support
- [x] Translation
- [x] Validation

## Breaking changes introduced

We will need to discuss these before merging.

#### Database

- To prevent unsafe model instantiating, the model constructors are now forced to only allow a single `$attributes` parameter via an interface (`\Winter\Storm\Database\ModelInterface` and `\Winter\Storm\Halcyon\ModelInterface`). This will ensure that calls like `Model::make()` or `Model::create()` will execute correctly. It is possible that some people might have used additional parameters for their model constructors - these will no longer work and must be moved to another method.
- Due to the above, Pivot model construction has been rewritten. The constructor used to allow 4 parameters but now only allows the one `$attributes` parameter as per the Model class. Construction now happens more closely to Laravel's format of calling a `Pivot::fromAttributes()` static method. If you previously used `new Pivot()` to create a pivot model, switch to using `Pivot::fromAttributes()` instead.
- The `MorphToMany` class now extends the `MorphToMany` class from Laravel, as opposed to the `BelongsToMany` class in Winter. This prevents repeated code in the Winter `MorphToMany` class and maintains covariance with Laravel. This will mean that it will no longer inherit from the Winter `BelongsToMany` class. To allow for this, we have converted most of the overridden `BelongsToMany` functionality in Winter into a trait (`Concerns\BelongsOrMorphsToMany`). The `BelongsToMany` class now also uses this trait.
- The relation traits found in `src/Database/Relations` have been moved to `src/Database/Relations/Concerns`, in order to keep just the actual relation classes within this directory. In the unlikely event that you are using a relation trait directly, please rename the trait classes to the following:
    - `Winter\Storm\Database\Relations\AttachOneOrMany` to `Winter\Storm\Database\Relations\Concerns\AttachOneOrMany`
    - `Winter\Storm\Database\Relations\DeferOneOrMany` to `Winter\Storm\Database\Relations\Concerns\DeferOneOrMany`
    - `Winter\Storm\Database\Relations\DefinedConstraints` to `Winter\Storm\Database\Relations\Concerns\DefinedConstraints`
    - `Winter\Storm\Database\Relations\HasOneOrMany` to `Winter\Storm\Database\Relations\Concerns\HasOneOrMany`
    - `Winter\Storm\Database\Relations\MorphOneOrMany` to `Winter\Storm\Database\Relations\Concerns\MorphOneOrMany`

#### Extension

- Previously, the `Winter\Storm\Extension\Extendable::extendClassWith()` method returned the current class if the extension name provided was an empty string. This appears to be a code smell, so this has been changed to throw an Exception.

#### Filesystem

- The `Filesystem::symbolizePath` method's `$default` parameter now accepts a `string`, `bool` or `null`. Only in the case of `null` will the method return the original path - the given `$default` will be used in all other cases. This is the same as the original functionality, but we're documenting it in case you have customised this method in some fashion.
- The `Filesystem::isPathSymbol` method previously returned the path symbol used, as a string, if one was found and `false` if not found. However, the docblock stipulated, as well as the method name itself implied, that this is a boolean check function, so we have converted this to a straight boolean response - `true` if a path symbol is used, otherwise `false`.
- Many classes within the `Filesystem` namespace have had type hinting and return types added to enforce functionality and clarify documentation. If you extend any of these classes in your plugins, you may need to update your method signatures.

#### Foundation

- The `Application` class callback methods `before` and `after` were documented as `void` methods, but returned a value. These no longer return a value.

#### Halcyon

- To prevent unsafe model instantiating, the model constructors are now forced to only allow a single `$attributes` parameter via an interface (`\Winter\Storm\Database\ModelInterface` and `\Winter\Storm\Halcyon\ModelInterface`). This will ensure that calls like `Model::make()` or `Model::create()` will execute correctly. It is possible that some people might have used additional parameters for their model constructors - these will no longer work and must be moved to another method.
- The Halcyon Builder `insert` method now returns an `int` representing the created model's filesize, not a `bool`.
- The Halcyon Builder `insert` method requires the `$values` parameter to actually contain variables and will throw an Exception if an empty array is provided. Previously, this was silently discarded and returned `true` (although this would not actually save the file).

#### HTML

- The `Winter\Storm\Html\FormBuilder` class has had type hinting and return types added to enforce functionality and clarify documentation. If you extend this class in your plugin, you may need to make changes to your method signatures if you overwrite the base functionality.

#### Parse
- The `Bracket` class constructor is now `final` to prevent unsafe static calls to the `parse` static method.
- The `Bracket::parseString()` method previously returned `false` if a string was not provided for parsing, or the string was empty after trimming. Since the signature calls for a string, we won't check for this anymore. If the string is empty, an empty string will be returned.
- The `markdown.beforeParse` event in the `Markdown` class sent a single `MarkdownData` instance as a parameter to the listeners. It now sends an array with the `MarkdownData` instance as the only value, to meet the signature requirements for events.
- The `Syntax\FieldParser` class constructor has been made `final` to prevent unsafe static calls to the `parse` statuc method.
- The `$template` parameter for the `Syntax\FieldParser` class constructor was originally optional. Since there is no purpose for this, and no way to populate the template after the fact, this has been made required.
- The `Syntax\Parser` class constructor has been made `final` to prevent unsafe static calss to the `parse` static method.
- The `$template` parameter for the `Syntax\Parser` class constructor was originally optional. Since there is no purpose for this, and no way to populate the template after the fact, this has been made required.

#### Testing
- The `Winter\Storm\Support\Testing\Fakes\MailFake::queue()` method has had its signature re-arranged to remain compatible with Laravel's `MailFake` class, with the `$queue` parameter being moved to the second parameter. The new signature is `($view, $queue, $data, $callback)`.


src/Database/Behaviors/Purgeable.php
```diff
@@ -4,10 +4,14 @@ class Purgeable extends \Winter\Storm\Extension\ExtensionBase
 {
     /**
      * @var array List of attribute names which should not be saved to the database.
-     *
-     * public $purgeable = [];
      */
+    public $purgeable = [];
```
LukeTowers: Why was this changed?

composer.json
```diff
@@ -52,7 +52,9 @@
         "squizlabs/php_codesniffer": "3.*",
         "php-parallel-lint/php-parallel-lint": "^1.0",
         "meyfa/phpunit-assert-gd": "^2.0.0|^3.0.0",
-        "dms/phpunit-arraysubset-asserts": "^0.1.0|^0.2.1"
+        "dms/phpunit-arraysubset-asserts": "^0.1.0|^0.2.1",
+        "nunomaduro/larastan": "^1.0",
+        "orchestra/testbench": "dev-master"
```
LukeTowers: Why was testbench added here?

phpstan.neon
```diff
@@ -0,0 +1,9 @@
+includes:
+    - ./vendor/nunomaduro/larastan/extension.neon
+
+parameters:
+    paths:
+        - src
+    level: 5
+    excludePaths:
+        - src/Assetic/*
```
LukeTowers: We need to get rid of the custom usage of Assetic in the first place

src/Halcyon/Datasource/Datasource.php
```diff
@@ -1,56 +1,91 @@
...
+    abstract public function lastModified(string $dirName, string $fileName, string $extension): ?int;
+
+    /**
+     * @inheritDoc
      */
-    public function makeCacheKey($name = '')
+    public function makeCacheKey(string $name = ''): string
     {
-        return crc32($name);
+        return hash('crc32b', $name);
```
LukeTowers, line 79: What's the reasoning behind changing the hash algorithm?

src/Halcyon/Datasource/DbDatasource.php
```diff
@@ -90,27 +86,22 @@ public function getQuery($ignoreDeleted = true)
     }
 
     /**
-     * Helper to make file path.
+     * Helper method to make the full file path to the model.
```
LukeTowers: "the full" implies the absolute path to the file on the filesystem, which is incorrect for databases and may not be correct for the filesystem source either (haven't looked). Thoughts?

src/Halcyon/Datasource/DbDatasource.php
```diff
@@ -372,52 +337,31 @@ public function delete(string $dirName, string $fileName, string $extension)
...
-     * Generate a cache key unique to this datasource.
-     *
-     * @param  string  $name
-     * @return string
+     * @inheritDoc
      */
-    public function makeCacheKey($name = '')
-    {
-        return crc32($this->source . $name);
-    }
```
LukeTowers, line 403: Why was this method removed?

src/Halcyon/Datasource/FileDatasource.php
```diff
@@ -348,54 +299,38 @@ protected function makeDirectoryPath($dirName, $relativePath = '')
...
-    /**
-     * Generate a cache key unique to this datasource.
-     *
-     * @param  string  $name
-     * @return string
-     */
-    public function makeCacheKey($name = '')
-    {
-        return crc32($this->basePath . $name);
-    }
```
LukeTowers, line 372: Why was this method removed?

src/Halcyon/Model.php
```diff
@@ -681,7 +682,8 @@ public function getAttribute($key)
         /**
          * @see Winter\Storm\Database\Model::getAttributeValue
          */
-        if (($attr = $this->fireEvent('model.beforeGetAttribute', [$key], true)) !== null) {
+        $attr = $this->fireEvent('model.beforeGetAttribute', [$key], true);
+        if ($attr !== null) {
```
LukeTowers:
```suggestion
        if (!is_null($attr)) {
```

src/Halcyon/Model.php
```diff
@@ -697,7 +699,8 @@ public function getAttribute($key)
         /**
          * @see Winter\Storm\Database\Model::getAttributeValue
          */
-        if (($_attr = $this->fireEvent('model.getAttribute', [$key, $value], true)) !== null) {
+        $_attr = $this->fireEvent('model.getAttribute', [$key, $value], true);
+        if ($_attr !== null) {
```
LukeTowers:
```suggestion
        if (!is_null($_attr)) {
```

src/Halcyon/Model.php
```diff
@@ -980,12 +983,12 @@ public function delete()
      */
     protected function performDeleteOnModel()
     {
-        $this->newQuery()->delete($this->fileName);
+        $this->newQuery()->delete();
```
LukeTowers, line 984: Why was this changed?

src/Halcyon/Model.php
```diff
@@ -1553,9 +1557,13 @@ public static function initCacheItem(&$item)
      */
     public static function flushDuplicateCache()
     {
-        if (MemoryCacheManager::isEnabled() && self::getCacheManager() !== null) {
-            self::getCacheManager()->driver()->flushInternalCache();
+        if (!MemoryCacheManager::isEnabled() || self::getCacheManager() === null) {
```
LukeTowers:
```suggestion
        if (!MemoryCacheManager::isEnabled() || is_null(self::getCacheManager())) {
```

src/Halcyon/Processors/SectionParser.php
```diff
@@ -182,9 +184,9 @@ public static function parseOffset($content)
      * Returns the line number of a found instance of CMS object section separator (==).
      * @param string $content Object content
      * @param int $instance Which instance to look for
-     * @return int The line number the instance was found.
+     * @return int|null The line number the instance was found.
      */
-    private static function calculateLinePosition($content, $instance = 1)
+    public static function calculateLinePosition($content, $instance = 1)
```
LukeTowers: Why was this changed?

src/Halcyon/Processors/SectionParser.php
```diff
@@ -209,7 +211,7 @@ private static function calculateLinePosition($content, $instance = 1)
      * @param int $startLine The calculated starting line from calculateLinePosition()
      * @return int The adjusted line number.
      */
-    private static function adjustLinePosition($content, $startLine = -1)
+    public static function adjustLinePosition($content, $startLine = -1)
```
LukeTowers: Why was this changed?

src/Support/Traits/Emitter.php
```diff
@@ -139,7 +139,8 @@ public function unbindEvent($event = null)
      * @param string $event Event name
      * @param array $params Event parameters
      * @param boolean $halt Halt after first non-null result
-     * @return array Collection of event results / Or single result (if halted)
+     * @return array|mixed|null Collection of event results as an array if not halted, if not halted. Otherwise, the first non-null result, or null
+     *  if no listeners returned a result.
```
LukeTowers:
```suggestion
     * @return array|mixed|null If halted, the first non-null result. If not halted, an array of event results. Returns 
     *  null if no listeners returned a result.
```

composer.json
bennothommo: It's needed for testing Laravel modules and not "true" applications - which Storm kinda sorta is not. I found I needed it to even get PHPStan working in the first place.

phpstan.neon
bennothommo: Indeed, but this is in there for now until that happens.

src/Database/Behaviors/Purgeable.php
bennothommo: It was being reported as a missing variable otherwise, but yeah I realise this is probably incorrect since purgeable needs to exist on the model. Will adjust.

src/Halcyon/Datasource/Datasource.php
bennothommo: It was the recommendation from the [`crc32()` method docs](https://www.php.net/crc32) if you needed hashes as strings.

src/Halcyon/Datasource/DbDatasource.php
bennothommo: Perhaps "Helper method to make the full relative file path to the model"? I just thought the original comment was not descriptive enough.

src/Halcyon/Datasource/DbDatasource.php
bennothommo: The `makeCacheKey` method is being provided by the abstract class `Winter\Storm\Halycon\Datasource\Datasource`. While this implementation of the method does use `$this->source`, the `getPathsCacheKey()` method has the source as part of the cache key prefix anyway, so it should still be unique. Is there any other reason why the source was necessary here?

src/Halcyon/Datasource/FileDatasource.php
bennothommo: Same reason as this: https://github.com/wintercms/storm/pull/64/files/c99be36471e01ac8ef93e5c0258a514ac850b866#r771805382

src/Halcyon/Model.php
bennothommo: The delete method on the model doesn't accept any parameters.

src/Halcyon/Processors/SectionParser.php
bennothommo: PHPStan reports this as unsafe private static call, since it's being called statically within the class. Realistically, I think this should probably be changed to a singleton.

src/Halcyon/Processors/SectionParser.php
bennothommo: Same as above.

phpstan.neon
LukeTowers: @bennothommo we can now get rid of this, although I'm not opposed to integrating Assetic with phpstan as well on its own repo.

src/Config/LoaderInterface.php
```diff
@@ -31,6 +31,15 @@ public function exists($group, $namespace = null);
...
     public function addNamespace($namespace, $hint);
 
+    /**
+     * Registers an alias for a given namespace.
+     *
+     * @param string $namespace
+     * @param string $alias
+     * @return void
+     */
+    public function registerNamespaceAlias(string $namespace, string $alias);
```
LukeTowers, line 41: Why was this required?

src/Config/Repository.php
```diff
@@ -115,6 +116,27 @@ public function get($key, $default = null)
...
+        foreach ($keys as $key => $default) {
+            if (is_numeric($key)) {
+                [$key, $default] = [$default, null];
+            }
+
+            $config[$key] = $this->get($key, $default);
+        }
+
+        return $config;
+    }
```
LukeTowers, line 138: Why was this added?

src/Config/Repository.php
```diff
@@ -148,48 +170,6 @@ public function set($key, $value = null)
...
-    /**
-     * Get all of the configuration items for the application.
-     *
-     * @return array
-     */
-    public function all()
-    {
-        return $this->items;
-    }
-
```
LukeTowers, line 192: Why were these removed?

src/Database/ModelInterface.php
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * The Model interface defines required methods for models to work.
+ *
+ * At the moment, this only restricts the signature for the constructor to only allow attributes as a parameter, in
+ * order to allow several static calls to work within the model architecture, but still allow models to extend the
+ * constructor if they wish.
+ *
+ * @author Winter CMS
+ */
+interface ModelInterface
```
LukeTowers, line 12: Why was this added?

src/Extension/ExtendableTrait.php
```diff
@@ -214,13 +214,16 @@ public function addDynamicProperty($dynamicName, $value = null)
 
     /**
      * Dynamically extend a class with a specified behavior
-     * @param  string $extensionName
+     * @param string $extensionName
      * @return void
      */
     public function extendClassWith($extensionName)
     {
         if (!strlen($extensionName)) {
```
LukeTowers: This should probably be changed to a `!empty()` check and the typehints moved into the method signature directly.

src/Filesystem/Definitions.php
```diff
@@ -12,22 +12,27 @@
 class Definitions
 {
 
+    /**
+     * Lock down the constructor for this class.
+     */
+    final public function __construct()
+    {
+    }
```
LukeTowers, line 20: Why is this required?

src/Filesystem/Definitions.php
```diff
@@ -37,13 +42,13 @@ public function getDefinitions($type)
     }
 
     /**
-     * Determines if a path should be ignored, sourced from the ignoreFiles
-     * and ignorePatterns definitions.
+     * Determines if a path should be ignored based on the ignoreFiles and ignorePatterns definitions.
+     *
+     * Returns `true` if the path is visible, `false` otherwise.
```
LukeTowers: This method name and comment is throwing me for a loop, they appear to be saying opposite things?

src/Filesystem/Filesystem.php
```diff
@@ -141,34 +144,33 @@ public function isLocalPath($path, $realpath = true)
     }
 
     /**
-     * Returns true if the provided disk is using the "local" driver
+     * Determines if the given disk is using the "local" driver.
      *
-     * @param Illuminate\Filesystem\FilesystemAdapter $disk
+     * @param \Illuminate\Filesystem\FilesystemAdapter $disk
      * @return boolean
```
LukeTowers: I don't see a reason to include docblock typehinting when we have method signature type hints setup; especially when the docblock ones don't add any additional information.

src/Filesystem/Zip.php
```diff
@@ -50,22 +50,23 @@
...
     /**
-     * Extract an existing zip file.
-     * @param  string $source Path for the existing zip
-     * @param  string $destination Path to extract the zip files
-     * @param  array  $options
-     * @return bool
+     * Lock down the constructor for this class.
      */
-    public static function extract($source, $destination, $options = [])
+    final public function __construct()
```
LukeTowers, line 60: Why is this required?

src/Config/LoaderInterface.php
bennothommo: I felt it made sense to have it - not that it's likely, but if anyone creates another type of Config loader, they'll need to support aliases.

src/Config/Repository.php
bennothommo: The `Repository` class now extends Laravel's `Repository` class, but to support aliases, we needed to adjust `getMany` to query through the `get` method - Laravel's original `getMany` just simply extracts from an array.

src/Config/Repository.php
bennothommo: As above, now that we extend Laravel's base `Repository` class, there's no need for these methods to be overridden.

src/Database/ModelInterface.php
bennothommo: It's in my notes in the PR, but basically, we are now enforcing a signature for the Model constructor that matches Laravel and prevent unsafe static calls. It's something that needs to be discussed, because it could *potentially* be a breaking change, but personally, I don't see a need for extra parameters in the constructor.

src/Extension/ExtendableTrait.php
bennothommo: After I've gone through all the PHPStan issues, the next step was to go through and convert the PHP docs to true typehints, so I do intend to do this at some point.

I'll change the `strlen` to `empty` though.

src/Filesystem/Definitions.php
bennothommo: Because `Definitions` is called statically, this ensures that static calls are safe even with extensions to the class.

src/Filesystem/Definitions.php
bennothommo: Your right - the comment is wrong. Will adjust :)

src/Filesystem/Filesystem.php
bennothommo: I'll leave the docblocks if they add additional context, but yeah, we can probably remove them if the typehints / return type are defined.

src/Filesystem/Zip.php
bennothommo: Protects against unsafe static calls.

src/Console/Command.php
```diff
@@ -76,7 +76,7 @@ public function complete(CompletionInput $input, CompletionSuggestions $suggesti
                     $dataType = 'Argument';
                     $suggestionType = 'Values';
                     break;
-                case 'options':
+                default:
```
LukeTowers: Why would we want to do this? An invalid case as written right now isn't possible; and if there was one I would rather it crashes than it continues silently. I didn't provide a default case on purpose 😉

bennothommo: Your logic makes PHPStan cry. It complains that the `$dataType` and `$suggestionType` vars *may* not exist.

bennothommo: @LukeTowers how about this instead? https://github.com/wintercms/storm/commit/35b36b7d686d82bc0eeead5fee1d8f3ee55bbb21

Keeps your `switch` call as is, and appeases PHPStan.

LukeTowers: But still fails silently. Could do a default that throws an exception I guess.

src/Halcyon/Processors/SectionParser.php
```diff
@@ -284,7 +284,7 @@ private static function calculateLinePosition(string $content, int $instance = 1
             }
 
             if ($count === $instance) {
-                return static::adjustLinePosition($content, $number);
+                return self::adjustLinePosition($content, $number);
```
LukeTowers: instead of changing this to self the method should be changed to protected and this should remain static. The other private method should be made protected too.

bennothommo:
I believe I have done all that I can with the code analysis. I've gotten the amount of errors from 1100 to ~130 unique errors, with nearly all of the remaining ones being caused by Eloquent magic that we really can't do much about.

I'm hesitant to introduce this as a test - I think if anything we can treat this as a cleanup and making us more in sync with Laravel. If we wanted to make this a test, we can set a baseline in PHPStan, which basically ignores all remaining errors and will only report on any *new* errors going forward.

At this stage, it's ready to review. I've tested this with the main repo's tests and it's not reporting any errors there, but I haven't done a heap of testing directly in Winter so we'll probably need to have it running for a bit on a real project to ensure everything works as is.

src/Argon/Argon.php
```diff
@@ -1,11 +1,12 @@
 <?php namespace Winter\Storm\Argon;
 
 use Carbon\Carbon as DateBase;
```
LukeTowers: On line 42 wouldn't it be more accurate to state that the return type is Argon, not DataBase? We could then get rid of the DateBase import.

LukeTowers:
@bennothommo great, looking forward to reviewing this! Before I do, are there any of the breaking changes that we can avoid making? If the breaking changes are just being made to satisfy the tool's need for pedantics then perhaps we can just ignore them as the baseline. I only want this PR to introduce breaking changes if they are absolutely essential and should have been made anyways.

src/Database/Attach/File.php
```diff
@@ -93,14 +98,10 @@ class File extends Model
...
      * Creates a file object from a file an uploaded file.
-     * @param Symfony\Component\HttpFoundation\File\UploadedFile $uploadedFile
+     * @param \Symfony\Component\HttpFoundation\File\UploadedFile $uploadedFile
      */
     public function fromPost($uploadedFile)
     {
-        if ($uploadedFile === null) {
-            return;
-        }
-
```
LukeTowers, line 103: What was the reason for this change?

src/Database/Attach/File.php
```diff
@@ -141,17 +142,13 @@ public function fromFile($filePath)
...
+     * @param string $filename Filename
      *
      * @return $this
      */
     public function fromData($data, $filename)
     {
-        if ($data === null) {
-            return;
-        }
-
```
LukeTowers, line 154: What was the reason for this change?

src/Database/Attach/File.php
```diff
@@ -786,10 +794,9 @@ protected function putFile($sourcePath, $destinationFileName = null)
          */
         if (
             !FileHelper::isDirectory($destinationPath) &&
-            !FileHelper::makeDirectory($destinationPath, 0777, true, true) &&
-            !FileHelper::isDirectory($destinationPath)
+            !FileHelper::makeDirectory($destinationPath, 0777, true, true)
```
LukeTowers, line 873: What was the reason for this change?

src/Database/Attach/File.php
```diff
@@ -866,14 +873,13 @@ protected function deleteEmptyDirectory($dir = null)
...
+     * @param string $dir The path to the directory.
+     *
+     * @return bool|null
      */
     protected function isDirectoryEmpty($dir)
     {
-        if (!$dir) {
-            return null;
-        }
-
```
LukeTowers, line 876: What was the reason for this change? It conflicts with the docblock return type now

src/Database/Behaviors/Purgeable.php
```diff
@@ -1,13 +1,16 @@
 <?php namespace Winter\Storm\Database\Behaviors;
 
+/**
+ * @deprecated 1.2.0. We recommend using the \Winter\Storm\Database\Traits\Purgeable trait instead.
+ */
```
LukeTowers: Why?

src/Database/Behaviors/Sortable.php
```diff
@@ -23,6 +23,7 @@
  *
  *   const SORT_ORDER = 'my_sort_order';
  *
+ * @deprecated 1.2.0. We recommend using the \Winter\Storm\Database\Traits\Sortable trait instead.
```
LukeTowers: Why? The DB traits as behaviours are intended for use by plugins extending other plugins.

src/Database/Connections/Connection.php
```diff
@@ -50,14 +48,31 @@ public function logQuery($query, $bindings, $time = null)
...
     }
+
+    /**
+     * Fire the given event if possible.
+     *
+     * @param string $event
+     * @param array|object $attributes
+     * @return void
+     */
+    protected function fireEvent($event, $attributes = [])
```
LukeTowers: Is this a Laravel override? If not we can probably add the type hints direct to the signature instead of the docblock

src/Database/MorphPivot.php
```diff
@@ -0,0 +1,185 @@
...
+            $query->orWhere(function ($query) use ($segments) {
+                return $query->where($segments[0], $segments[1])
+                    ->where($segments[2], $segments[3])
+                    ->where($segments[4], $segments[5]);
+            });
+        }
+
+        return $query;
+    }
+}
```
LukeTowers, line 190: Do we have WIP documentation for this?

src/Database/Traits/Purgeable.php
```diff
@@ -54,26 +54,24 @@ public function addPurgeable($attributes = null)
...
+        } else {
             $purgeable = $this->getPurgeableAttributes();
         }
 
         $attributes = $this->getAttributes();
         $cleanAttributes = array_diff_key($attributes, array_flip($purgeable));
         $originalAttributes = array_diff_key($attributes, $cleanAttributes);
 
-        if (is_array($this->originalPurgeableValues)) {
+        if (count($this->originalPurgeableValues)) {
```
LukeTowers, line 72: Can this property ever be anything that isn't countable()? Why was this change needed?

src/Database/Traits/Sortable.php
```diff
@@ -75,6 +75,6 @@ public function setSortableOrder($itemIds, $itemOrders = null)
      */
     public function getSortOrderColumn()
     {
-        return defined('static::SORT_ORDER') ? static::SORT_ORDER : 'sort_order';
+        return defined('static::SORT_ORDER') ? constant('static::SORT_ORDER') : 'sort_order';
```
LukeTowers, line 78: Why?

src/Filesystem/Definitions.php
LukeTowers: Can you give me an example of unsafe behaviour that this protects against? Apologies if you already have, probably just forgot.

src/Foundation/Exception/Handler.php
```diff
@@ -176,14 +176,14 @@ protected function callCustomHandlers($throwable, $fromConsole = false)
...
             }
-            catch (Throwable $t) {
-                $response = $this->convertThrowableToResponse($t);
-            }
+
             // If this handler returns a "non-null" response, we will return it so it will
             // get sent back to the browsers. Once the handler returns a valid response
             // we will cease iterating through them and calling these other handlers.
-            if (isset($response) && ! is_null($response)) {
+            if (isset($response)) {
```
LukeTowers, line 186: Why was this required?

LukeTowers (commented):
Some initial comments before I head to bed, looking good so far @bennothommo!

bennothommo:
@LukeTowers I feel all of the BC breaks above will be necessary, as they either fix broken functionality, move us closer to Laravel or are documentation changes. There might be some small tweaks outside of those that may not be necessary and were just to appease PHPStan, but at this point, I think it all is in service of making sure our documentation is accurate and our functionality works smoothly. The idea of setting the baseline is that we have as few issues as possible, because otherwise there's a lot more chance for errors to be reported in the future.

src/Database/Attach/File.php
bennothommo: `$uploadedFile` should not be null, according to both the docblock and the signature. If it is, this code block is silently discarding it, when it should throw an exception.

src/Database/Attach/File.php
bennothommo: Same reason as the `fromPost` changes.

src/Database/Attach/File.php
bennothommo: The second check to `isDirectory` is redundant. If `makeDirectory` fails, it's already returning `false` which triggers the error.

src/Database/Attach/File.php
```diff
@@ -866,14 +873,13 @@ protected function deleteEmptyDirectory($dir = null)
 
     /**
      * Returns true if a directory contains no files.
-     * @return void
+     *
+     * @param string $dir The path to the directory.
+     *
+     * @return bool|null
```
bennothommo:
```suggestion
     * @return bool
```

src/Database/Attach/File.php
bennothommo: You're right. Fixed the docblock.

src/Database/Behaviors/Purgeable.php
bennothommo: I'm not really sure why anymore. I think it was when I was railing against behaviours, but you're right, they have their use. We should probably make this share functionality with the trait if possible though, so we don't have to modify code in two places.

src/Database/Connections/Connection.php
bennothommo: It's not, no. I'll add the type hints.

src/Database/MorphPivot.php
bennothommo: Not yet, I still need to add some unit tests and make sure it works.

src/Database/Traits/Purgeable.php
bennothommo: No, it'll always be an array.

src/Database/Traits/Sortable.php
bennothommo: Both PHPStan AND Intellisense/Intelephense report it as a missing constant, even with the `defined` check surrounding it.

src/Filesystem/Definitions.php
bennothommo: It probably will never be a problem for this class in particular, but since this change has been done to a few classes, I'll try and explain the situation better.

So with classes where we have static calls that create a new instance of the same class, like the `get` method in this class:

```php
    public static function get(string $type): array
    {
        return (new static)->getDefinitions($type);
    }
```

This method will break if someone were to extend the class and add an additional parameter to the constructor, because, as you see above, the method is returning a new instance of the class *without* any parameters. Basically, our static calls are expecting a certain signature for the constructor, which an extending class might break.

Making a constructor final prevents this, by enforcing a certain signature for the constructor, ensuring that the `new static` calls will always work.

src/Foundation/Exception/Handler.php
bennothommo: `isset` and `!is_null` is the same test.

src/Database/Traits/Sortable.php
LukeTowers: Silly automated tools...

src/Database/Behaviors/Purgeable.php
LukeTowers: I think we tried when we first added it and it wasn't that easy to do.

src/Database/Behaviors/Sortable.php
mjauvin: Yeah, I added this to dynamically make a model Sortable when extending a plugin.

src/Console/Command.php
```diff
@@ -97,6 +97,8 @@ public function complete(CompletionInput $input, CompletionSuggestions $suggesti
                     $dataType = 'Option';
                     $suggestionType = 'Options';
                     break;
+                default:
+                    throw new \Exception('Invalid input type being parsed during completion');
```
LukeTowers, line 102:
```suggestion
                    // This should not be possible to ever be triggered given the type is hardcoded above
                    throw new \Exception('Invalid input type being parsed during completion');
```

src/Database/Attach/File.php
```diff
@@ -93,14 +98,10 @@ class File extends Model
 
     /**
      * Creates a file object from a file an uploaded file.
-     * @param Symfony\Component\HttpFoundation\File\UploadedFile $uploadedFile
+     * @param \Symfony\Component\HttpFoundation\File\UploadedFile $uploadedFile
      */
     public function fromPost($uploadedFile)
```
LukeTowers, line 106: Should we type hint the parameter now to make the error more explicit when incorrect data is provided rather than being "Failed to access method getClientOriginalName() on null"?

src/Database/Attach/File.php
```diff
@@ -141,17 +142,13 @@ public function fromFile($filePath)
...
+     * @param string $data Raw data
+     * @param string $filename Filename
      *
      * @return $this
      */
     public function fromData($data, $filename)
     {
-        if ($data === null) {
-            return;
-        }
```
LukeTowers, line 153: Likewise, should we add the type hint to the method signature now?

src/Database/Model.php
```diff
@@ -529,7 +529,7 @@ public function getObservableEvents()
     /**
      * Get a fresh timestamp for the model.
      *
-     * @return \Winter\Storm\Argon\Argon
+     * @return \Illuminate\Support\Carbon
```
LukeTowers, line 535: Why change this? Isn't Argon more specific than Carbon?

src/Database/Models/Revision.php
```diff
@@ -34,7 +34,7 @@ public function getNewValueAttribute($value)
      */
     public function getOldValueAttribute($value)
     {
-        if ($this->cast == 'date' && !is_null($value)) {
+        if ($this->getAttribute('cast') === 'date' && !is_null($value)) {
```
LukeTowers, line 36: Why was this required?

src/Database/MorphPivot.php
LukeTowers: Any progress on docs and unit tests? Can we copy the DB test from Laravel for this?

src/Database/MorphPivot.php
```diff
@@ -0,0 +1,185 @@
+<?php namespace Winter\Storm\Database;
+
+class MorphPivot extends Pivot
```
LukeTowers, line 8:
```suggestion
/**
 * Reimplementation of \Illuminate\Database\Eloquent\Relations\MorphPivot
 * but extends Winter's Pivot base class instead of Laravel's to get access to
 * Winter magic.
 */
class MorphPivot extends Pivot
```

src/Database/Pivot.php
```diff
@@ -1,17 +1,24 @@
...
 
 class Pivot extends Model
 {
     /**
      * The parent model of the relationship.
      *
-     * @var \Illuminate\Database\Eloquent\Model
+     * @var \Illuminate\Database\Eloquent\Model|null
      */
     protected $parent;
```
LukeTowers: Why was this property retained instead of replaced as I did in https://github.com/wintercms/storm/pull/65?

src/Database/Pivot.php
```diff
@@ -1,17 +1,24 @@
 <?php namespace Winter\Storm\Database;
 
 use Illuminate\Database\Eloquent\Model as ModelBase;
-use Illuminate\Database\Eloquent\Builder as BuilderBase;
+use Winter\Storm\Support\Str;
 
 class Pivot extends Model
```
LukeTowers: Instead of replicating a the AsPivot trait in our code, can we just implement it in our Pivot model and then override any methods that need tweaking?

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php

LukeTowers (changes requested):

src/Database/Model.php
bennothommo: It matches Laravel's signature, I believe.

src/Database/Models/Revision.php
bennothommo: It fixes an issue where the `cast` property might not be set.

src/Database/Pivot.php
bennothommo: I couldn't discount that `$parent` might be used by our users. The edits I made to the Pivot keep it backwards compatible.

src/Database/Models/Revision.php
LukeTowers: Wouldn't that just cycle through the getAttribute() logic anyways by way of the `__get()` magic method?

src/Database/MorphPivot.php
```diff
@@ -0,0 +1,190 @@
+<?php namespace Winter\Storm\Database;
+
+/**
+ * Reimplementation of \Illuminate\Database\Eloquent\Relations\MorphPivot
+ * but extends Winter's Pivot base class instead of Laravel's to get access to
+ * Winter magic.
+ */
+class MorphPivot extends Pivot
```
LukeTowers, line 8: @bennothommo I've added a [comment](https://github.com/laravel/framework/pull/25851#issuecomment-1133820077) to the original AsPivot PR making a case for the addition of an AsMorphPivot trait, we could potentially submit the PR ourselves to simplify our codebase.

src/Database/Pivot.php
LukeTowers: @bennothommo we should be able to use the AsPivot trait instead in our model and just handle the BC for the parent property by defining getParentAttribute() & setParentAttribute() methods. Thoughts?

src/Database/QueryBuilder.php
```diff
@@ -112,14 +113,10 @@ public function get($columns = ['*'])
...
      *
      * @param  array  $columns
-     * @return array
+     * @return BaseCollection
      */
     protected function getDuplicateCached($columns = ['*'])
     {
-        if (is_null($this->columns)) {
-            $this->columns = $columns;
-        }
```
LukeTowers, line 121: Why was this removed?

src/Database/QueryBuilder.php
```diff
@@ -140,14 +137,10 @@ protected function getDuplicateCached($columns = ['*'])
...
      *
      * @param  array  $columns
-     * @return array
+     * @return BaseCollection
      */
     public function getCached($columns = ['*'])
     {
-        if (is_null($this->columns)) {
-            $this->columns = $columns;
-        }
```
LukeTowers, line 149: Why was this removed?

src/Database/Relations/AttachMany.php
```diff
@@ -81,7 +80,7 @@ public function getSimpleValue()
 
         if ($files) {
             $value = [];
-            foreach ($value as $file) {
+            foreach ($files as $file) {
```
LukeTowers, line 86: Does this cause any side effects having this unreported bug fixed?

src/Database/Behaviors/Purgeable.php
LukeTowers:
```suggestion
```

src/Database/Behaviors/Sortable.php
LukeTowers:
```suggestion
```

src/Database/Relations/AttachOneOrMany.php
```diff
@@ -1,303 +1,9 @@
 <?php namespace Winter\Storm\Database\Relations;
 
-use Illuminate\Database\Eloquent\Model;
-use Illuminate\Database\Eloquent\Builder;
-use Winter\Storm\Support\Facades\DbDongle;
-use Winter\Storm\Database\Attach\File as FileModel;
-use Symfony\Component\HttpFoundation\File\UploadedFile;
-
+/**
+ * @deprecated 1.2.0 Use `\Winter\Storm\Database\Relations\Concerns\AttachOneOrMany` instead.
```
LukeTowers: Can you explain more as to what's the motivation behind moving all these relation classes to a new location? What's the benefit?

src/Extension/ExtendableTrait.php
```diff
@@ -1,12 +1,12 @@
 <?php namespace Winter\Storm\Extension;
 
-use App;
 use Exception;
 use ReflectionClass;
 use ReflectionMethod;
 use BadMethodCallException;
 use Winter\Storm\Support\ClassLoader;
 use Winter\Storm\Support\Serialization;
+use Illuminate\Support\Facades\App;
```
LukeTowers, line 9: Why do some files need to have the short form import removed but others don't? What's the point?

src/Filesystem/Filesystem.php
```diff
@@ -141,34 +144,30 @@ public function isLocalPath($path, $realpath = true)
...
 
     /**
-     * Determine if a file exists with case insensitivity
-     * supported for the file only.
-     * @param  string $path
-     * @return mixed  Sensitive path or false
+     * Determines if a file exists (ignoring the case for the filename only).
      */
-    public function existsInsensitive($path)
+    public function existsInsensitive(string $path): string|false
```
LukeTowers, line 170: Is `false` a valid return type in PHP 8?

src/Filesystem/Filesystem.php
```diff
@@ -141,34 +144,30 @@ public function isLocalPath($path, $realpath = true)
...
-     * @param  mixed  $className Class name or object
-     * @return string The file path
+     * Finds the path of a given class.
+     *
+     * Returns `false` if the path cannot be determined.
+     *
+     * @param string|object $className Class name or object
      */
-    public function fromClass($className)
+    public function fromClass(string|object $className): string|false
```
LukeTowers, line 161: Is `false` a valid return type in PHP 8?

src/Filesystem/Filesystem.php
```diff
@@ -191,52 +190,48 @@ public function existsInsensitive($path)
...
+     *
+     * Returns the original path if no symbol is used, and no default is specified.
      */
-    public function symbolizePath($path, $default = false)
+    public function symbolizePath(string $path, string|bool|null $default = null): string
     {
-        if (!$firstChar = $this->isPathSymbol($path)) {
-            return $default === false ? $path : $default;
+        if (!$this->isPathSymbol($path)) {
+            return (is_null($default)) ? $path : $default;
```
LukeTowers, line 208: Why is this required? Wouldn't this mean that the return type cannot be only string since passing a path with no path symbol and a `$default` value of `false` or `true` will result in a bool being returned instead of a string? How come static analysis didn't catch this?

src/Filesystem/Filesystem.php
```diff
@@ -191,52 +190,48 @@ public function existsInsensitive($path)
...
-     * @param  string  $path
-     * @return boolean
+     * Determines if the given path is using a path symbol.
      */
-    public function isPathSymbol($path)
+    public function isPathSymbol(string $path): bool
     {
-        $firstChar = substr($path, 0, 1);
-        if (isset($this->pathSymbols[$firstChar])) {
-            return $firstChar;
```
LukeTowers, line 229: Any concerns about this breaking change?

src/Filesystem/Filesystem.php
```diff
@@ -411,10 +405,8 @@ public function fileNameMatch($fileName, $pattern)
 
     /**
      * Finds symlinks within the base path and provides a source => target array of symlinks.
```
LukeTowers:
```suggestion
     * Finds symlinks within the base path and populates the local symlinks property with an array of source => target symlinks.
```

src/Filesystem/Zip.php
```diff
@@ -50,22 +50,23 @@
...
     {
-        extract(array_merge([
-            'mask' => 0777
-        ], $options));
+    }
+
+    /**
+     * Extracts an existing ZIP file.
+     */
+    public static function extract(string $source, string $destination, array $options = []): bool
```
LukeTowers, line 72: Can we document the valid elements for the `$options` array? Looks like right now all that's supported is the `mask` element.

src/Filesystem/Zip.php
```diff
@@ -108,13 +105,13 @@ public static function make($destination, $source, $options = [])
...
 
     /**
-     * Includes a source to the Zip
-     * @param mixed $source
-     * @param array $options
-     * @return self
+     * Adds a source file or directory to a Zip file.
      */
-    public function add($source, $options = [])
+    public function add(string $source, array $options = []): self
```
LukeTowers, line 129: Can we document the valid options here as well?

src/Filesystem/Zip.php
```diff
@@ -124,23 +121,18 @@ public function add($source, $options = [])
...
         }
 
-        extract(array_merge([
-            'recursive' => true,
-            'includeHidden' => false,
-            'basedir' => dirname($source),
-            'baseglob' => basename($source)
-        ], $options));
+        $basedir = dirname($source);
+        $baseglob = basename($source);
```
LukeTowers: Isn't this a breaking change by no longer supporting `$basedir` and `$baseglob` in the options array?

src/Foundation/Application.php
```diff
@@ -390,12 +390,12 @@ public function setLocale($locale)
...
      * Register all of the configured providers.
      *
-     * @var bool $isRetry If true, this is a second attempt without the cached packages.
+     * @param bool $isRetry If true, this is a second attempt without the cached packages.
      * @return void
      */
     public function registerConfiguredProviders($isRetry = false)
     {
-        $providers = Collection::make($this->config['app.providers'])
+        $providers = Collection::make($this->get('config')['app.providers'])
```
LukeTowers, line 399: Why is this required?

src/Halcyon/ModelInterface.php
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * The Model interface defines required methods for models to work.
+ *
+ * At the moment, this only restricts the signature for the constructor to only allow attributes as a parameter, in
+ * order to allow several static calls to work within the model architecture, but still allow models to extend the
+ * constructor if they wish.
+ *
+ * @author Winter CMS
+ */
+interface ModelInterface
```
LukeTowers, line 12: Why wouldn't Halcyon use the Winter Eloquent model interface?

src/Halcyon/Processors/SectionParser.php
```diff
@@ -65,13 +65,14 @@ protected static function parseIntoSections(string $content, int $limit = 3): ar
...
             'wrapCodeInPhpTags' => true,
             'isCompoundObject'  => true,
-        ], $options));
+        ], $options);
+        extract($sectionOptions);
 
-        if (!$isCompoundObject) {
-            return array_get($data, 'content', '');
+        if (!isset($isCompoundObject) || $isCompoundObject === false) {
+            return array_get($data, 'content');
```
LukeTowers: Why did you remove the empty string as the default value here?

src/Html/BlockBuilder.php
```diff
@@ -75,30 +75,21 @@ public function endBlock($append = false)
...
-     *
-     * @param string $name Specifies the block name.
-     * @param string $content Specifies the block content.
-     * @return void
      */
-    public function append($name, $content)
+    public function append(string $name, string $content): void
     {
         if (!isset($this->blocks[$name])) {
             $this->blocks[$name] = '';
```
LukeTowers, line 95: Unrelated to the current PR, doesn't this mean that if you call append targeting a block that doesn't exist yet if you call it again a second time it will exist in the second instance but not the first?

src/Parse/Markdown.php
```diff
@@ -90,16 +105,19 @@ public function parseLine($text)
...
     {
+        if (is_null($parser)) {
+            $parser = $this->getParser();
+        }
         $data = new MarkdownData($text);
 
-        $this->fireEvent('beforeParse', $data, false);
-        Event::fire('markdown.beforeParse', $data, false);
+        $this->fireEvent('beforeParse', [$data], false);
+        Event::fire('markdown.beforeParse', [$data], false);
```
LukeTowers, line 120: Does this cause any issues having it changed to an array instead of just the variable?

src/Parse/Markdown.php
```diff
@@ -31,14 +31,37 @@ class Markdown
...
+    /**
+     * Gets an instance of the parser.
+     *
+     * We return a new instance each time to prevent contamination of clean instances.
+     *
+     * @return Parsedown
+     */
+    protected function getParser()
+    {
+        return new $this->parserClass;
```
LukeTowers, line 47: Why was this changed to no longer cache the parser class instance?

src/Support/Facade.php
```diff
@@ -10,31 +10,4 @@
...
-
-    /**
-     * If the accessor is not found via getFacadeAccessor, use this instance as a fallback.
-     *
-     * @return mixed
-     */
-    protected static function getFacadeInstance()
-    {
-        return null;
-    }
```
LukeTowers, line 39: Why were these changes required?

src/Support/Facades/Input.php
```diff
@@ -18,6 +21,18 @@ public static function get($key = null, $default = null)
...
+     * Gets all input data items.
+     *
+     * This method is used for all request verbs (GET, POST, PUT, and DELETE)
+     *
+     * @return array|null
+     */
+    public static function all()
+    {
+        return static::$app['request']->input();
+    }
```
LukeTowers, line 34: Why was this required?

src/Support/Serialization.php
```diff
@@ -12,11 +12,11 @@ class Serialization
      * Wraps a closure in a SerializableClosure, returns the provided object if it's not a closure.
      *
      * @param Closure|mixed $callable provided callable to be wrapped if it's a closure
-     * @return mixed|SerializableClosure
+     * @return SerializableClosure|mixed
      */
     public static function wrapClosure($callable)
     {
-        if ($callable instanceof Closure && !($callable instanceof SerializableClosure)) {
+        if ($callable instanceof Closure) {
```
LukeTowers, line 19: Won't this now double wrap SerializableClosures?

src/Validation/Factory.php
```diff
@@ -10,6 +10,13 @@
  */
 class Factory extends BaseFactory implements FactoryContract
 {
+    /**
+     * The Validator resolver instance.
+     *
+     * @var \Closure|null
+     */
+    protected $resolver;
```
LukeTowers: Why was this change required?

LukeTowers (changes requested):

src/Database/Models/Revision.php
bennothommo: It does, but where possible, we should avoid relying on magic in internal methods as it's not easily deducible. Userland code may use magic.

src/Database/Relations/AttachMany.php
bennothommo: Perhaps it would make attachments be displayed correctly in the Lists widget? That's it? :P

src/Database/QueryBuilder.php
bennothommo: I think this was due to PHPStan saying that `$this->columns` can never be null.

src/Database/QueryBuilder.php
bennothommo: Same answer as https://github.com/wintercms/storm/pull/64#discussion_r878865484

src/Database/Relations/AttachOneOrMany.php
bennothommo: Due to the naming of the classes, it was getting a bit too hard to track what was an actual relation vs. what was a trait when all these files were mixed in the same directory under `src/Database/Relations`. I thought a tidy up would be useful here, in making sure that the traits were in a concerns subdirectory, and only the relations were kept in the main directory.

At the same time, I realised people could've been using the traits directly, so I had to keep the stubs in place. I would say though that it would be beneficial to remove them in the next main release.

TL;DR: Just clean-up of the codebase.

src/Extension/ExtendableTrait.php
bennothommo: Are you meaning the full facade names? I'm replacing the short facade names or inline uses with the full facade names everywhere I see.

src/Filesystem/Filesystem.php
bennothommo: It is indeed.

image

src/Filesystem/Filesystem.php
bennothommo: Same answer as https://github.com/wintercms/storm/pull/64#discussion_r878866343

src/Filesystem/Filesystem.php
bennothommo: Not sure why it wasn't picked up? Maybe a higher level check?

src/Filesystem/Filesystem.php
bennothommo: I'll make a note of it, but given the docblock itself said it was *supposed* to be a boolean, and the name of the function implies a boolean check, I think it's very unlikely people are using it to extract the path symbol.

src/Database/Models/Revision.php
LukeTowers: Hmm, I'm not sure. It would change the behaviour if someone had specified a `cast` property on the model - in these sorts of cases I feel like it's better to use the standard way of interacting with model attributes rather than bypass one of the layers.

src/Database/Relations/AttachOneOrMany.php
LukeTowers: How likely is it that people were actually using these traits? What would be the use case for that? If we're doing it to cleanup and it's not that likely that they were actually being used then we might as well just remove them in this version, otherwise we're not really seeing the benefits of the cleanup.

src/Extension/ExtendableTrait.php
LukeTowers: Yes, what's the benefit in doing that?

src/Database/Attach/File.php
```diff
@@ -270,33 +258,34 @@ public function getSizeAttribute()
...
+     * @return \Illuminate\Http\Response|void
      */
-    public function output($disposition = 'inline', $returnResponse = false)
+    public function output(string $disposition = 'inline', bool $returnResponse = false)
     {
         $response = response($this->getContents())->withHeaders([
             'Content-type'        => $this->getContentType(),
             'Content-Disposition' => $disposition . '; filename="' . $this->file_name . '"',
-            'Cache-Control'       => 'private, no-store, no-cache, must-revalidate, pre-check=0, post-check=0, max-age=0',
+            'Cache-Control'       => 'private, no-store, no-cache, must-revalidate, max-age=0',
```
LukeTowers, line 298: What's the reason for this change?

src/Database/Attach/File.php
```diff
@@ -397,15 +385,15 @@ public function getContentType()
...
+    public function getContents(?string $fileName = null): string
     {
         return $this->storageCmd('get', $this->getDiskPath($fileName));
     }
 
     /**
      * Returns the public address to access the file.
      */
-    public function getPath($fileName = null)
+    public function getPath(?string $fileName = null): string
```
LukeTowers: Adding type hints to all of these methods means that we'll need to propagate them to the System\Models\File model as well as any userland extensions of that class.

src/Database/Behaviors/Purgeable.php
```diff
@@ -76,12 +76,7 @@ public function purgeAttributes($attributesToPurge = null)
         $cleanAttributes = array_diff_key($attributes, array_flip($purgeable));
         $originalAttributes = array_diff_key($attributes, $cleanAttributes);
 
-        if (is_array($this->originalPurgeableValues)) {
-            $this->originalPurgeableValues = array_merge($this->originalPurgeableValues, $originalAttributes);
-        }
-        else {
-            $this->originalPurgeableValues = $originalAttributes;
-        }
+        $this->originalPurgeableValues = array_merge($this->originalPurgeableValues, $originalAttributes);
```
LukeTowers, line 78: Why was this change made? Does this property always exist as at least an array? If so where is it defined?

src/Database/Relations/AttachOneOrMany.php
bennothommo: I don't think it would be very likely anyone was using the traits directly.

src/Extension/ExtendableTrait.php
bennothommo: I think we alias the short names in the System module, not the library, so when running static analysis or doing code edits directly on the library, it isn't able to determine the correct class from the short names, so this change makes it explicit.

src/Database/Attach/File.php
bennothommo: `pre-check` and `post-check` are invalid Cache Control attributes. Internet Explorer 5 introduced them, and they haven't been used for years. [Relevant blog post for more info](https://docs.microsoft.com/en-us/archive/blogs/ieinternals/internet-explorers-cache-control-extensions?_fsi=f81xrzF2)

src/Database/Relations/AttachOneOrMany.php
LukeTowers: Let's just document the change and remove the original traits then.

src/Extension/ExtendableTrait.php
LukeTowers: Gotcha. Can we do some find and replace to make it fully explicit everywhere in this repo then?

src/Database/Models/Revision.php
bennothommo: Sorry @LukeTowers, just circling back to this only now. I'm not sure what you mean here though - both using this `getAttribute` call or using the `__get` magic method do the same thing, so what behaviour would be changed here?

src/Filesystem/Zip.php
bennothommo: Re-added the ability to overwrite these two variables through the options.

src/Foundation/Application.php
bennothommo: Because PHPStan is unable to follow through the magic methods, like with the example before with model attributes. This makes it explicit what is happening in a way that PHPStan can follow.

src/Halcyon/ModelInterface.php
bennothommo: Just seemed to be cleaner to have its own interface - having it implement the database interface may be misconstrued as Halcyon operating off the databases.

src/Halcyon/Processors/SectionParser.php
bennothommo: Not sure, actually. I'll re-add it.

src/Html/BlockBuilder.php
bennothommo: Correct - if you try to append a block that doesn't exist, this acts more or less like the `set` method. The second time you run it, the block would exist and this would therefore append to the block.

src/Parse/Markdown.php
bennothommo: It is listed as one of the breaking changes. Since event payloads are processed into an array, this was done mainly to appease PHPStan, but it's probably something we should stick to with events anyway.

src/Parse/Markdown.php
bennothommo: It was to prevent cross-contamination of data with subsequent Markdown calls. Each Markdown call will now have a clean instance of the parser to work with.

src/Support/Facade.php
bennothommo: As far as I can tell, this method is not used anywhere in the code. ~~Seemed to be a holdover from a previous Laravel version.~~

EDIT: Actually, I think this was to bring us closer in line to Laravel.

src/Support/Facades/Input.php
bennothommo: Mainly for the `input()` helper method - it will try to retrieve all input if no name is provided to that method. This is synonymous with the `post()` and `get()` helper methods.

src/Support/Serialization.php
bennothommo: `SerializableClosure` is not an instance of `Closure`, so it should skip this condition block and simply be returned.

src/Validation/Factory.php
bennothommo: It was an attempt to appease PHPStan on another line - there is an `is_null` check for `$this->resolver` which PHPStan says will always be true because according to the docs for resolver, it can never be `null`. This is not strictly speaking correct, but it is how the API defines it.

It doesn't work, because now the doc is not covariant with Laravel's original docs, so I'll remove it and rely on Laravel's definition.

src/Database/Behaviors/Purgeable.php
bennothommo: It will always be an array, so the condition that was used previously is redundant.

Lines 68-70 above will convert a single string provided to the argument into an array, and the `purgeable` property itself is defined as an array.

src/Database/Models/Revision.php
LukeTowers: @bennothommo I mean that if someone defines `$cast` as a property on a class that extends this model class then this logic would no longer get that property value. However, this is unlikely so I no longer care about this change :P

src/Database/Relations/HasOne.php
```diff
@@ -22,21 +22,6 @@ public function __construct(Builder $query, Model $parent, $foreignKey, $localKe
...
-    public function getResults()
-    {
-        // New models have no possibility of having a relationship here
-        // so prevent the first orphaned relation from being used.
-        if (!$this->parent->exists) {
-            return null;
-        }
-
-        return parent::getResults();
-    }
```
LukeTowers, line 38: Left over code from a bygone era?

src/Html/FormBuilder.php
```diff
@@ -525,48 +458,33 @@ public function selectRange($name, $begin, $end, $selected = null, $options = []
...
-     * @return string
      */
-    public function selectMonth($name, $selected = null, $options = [], $format = '%B')
+    public function selectMonth(string $name, string|array|null $selected = null, array $options = [], $format = '%B'): string
     {
         $months = [];
 
         foreach (range(1, 12) as $month) {
-            $months[$month] = strftime($format, mktime(0, 0, 0, $month, 1));
+            $months[$month] = strftime($format, mktime(12, 0, 0, $month, 1));
```
LukeTowers: Why was this change needed?

src/Html/FormBuilder.php
```diff
@@ -936,7 +828,7 @@ protected function getAppendage($method)
      *
      * @param  string  $name
      * @param  array   $attributes
-     * @return string
+     * @return string|null
```
LukeTowers: Can this return null if you default to returning an empty string?

LukeTowers (changes requested):

src/Database/Attach/File.php
LukeTowers: @bennothommo ping

src/Parse/Markdown.php
LukeTowers: Does the parser contain dangerous state?

bennothommo: Not confirmed, but it was better to be safe. I don't expect people would use too many Markdown calls in a single request.

src/Database/Attach/File.php
bennothommo: What would you prefer? Roll them back or mark it as a breaking change? I'm erring towards marking it as breaking, since it would be better to have full type hinting, and I can't imagine too many people would've extended this model.

src/Database/MorphPivot.php
bennothommo: Laravel had barely any unit tests for MorphPivot in particular. I've added a couple.

src/Database/Relations/HasOne.php
bennothommo: Nah, the parent classes do have that method, and handle it better from what I can see, so just using that.

src/Html/FormBuilder.php
bennothommo: It's a quick fix that I deploy on most sites to ensure that timezones don't affect the result, but thinking about it, it's not really required here, so I'll revert.

src/Html/FormBuilder.php
bennothommo: I think I had interpreted that `null` could be a value within the `$attributes` array, and the `array_key_exists` condition could still pass. But if we go by that logic, then *anything* could be part of that attributes array, so I've changed it to encourage a `string` return.

src/Database/Model.php
```diff
@@ -803,15 +805,6 @@ protected function saveInternal($options = [])
...
 
-        /*
-         * If there is nothing to update, Eloquent will not fire afterSave(),
-         * events should still fire for consistency.
-         */
-        if ($result === null) {
-            $this->fireModelEvent('updated', false);
-            $this->fireModelEvent('saved', false);
-        }
-
```
LukeTowers, line 814: Why was this removed?

src/Database/Relations/Concerns/DeferOneOrMany.php
```diff
@@ -35,24 +38,27 @@ public function withDeferred($sessionKey)
                         $query
                             ->select($this->parent->getConnection()->raw(1))
                             ->from($this->table)
-                            ->where($this->getOtherKey(), DbDongle::raw(DbDongle::getTablePrefix().$this->related->getQualifiedKeyName()))
+                            ->where($this->getOtherKey(), DbDongle::raw(
+                                DbDongle::getTablePrefix().$this->related->getQualifiedKeyName()
```
LukeTowers:
```suggestion
                                DbDongle::getTablePrefix() . $this->related->getQualifiedKeyName()
```

src/Database/Relations/Concerns/DeferOneOrMany.php
```diff
@@ -35,24 +38,27 @@ public function withDeferred($sessionKey)
...
-                     * Custom query for BelongsToManyBase since a "join" cannot be used
+                     * Custom query for BelongsToMany since a "join" cannot be used
                      */
                     $query->whereExists(function ($query) {
                         $query
                             ->select($this->parent->getConnection()->raw(1))
                             ->from($this->table)
-                            ->where($this->getOtherKey(), DbDongle::raw(DbDongle::getTablePrefix().$this->related->getQualifiedKeyName()))
+                            ->where($this->getOtherKey(), DbDongle::raw(
+                                DbDongle::getTablePrefix().$this->related->getQualifiedKeyName()
```
LukeTowers:
```suggestion
                                DbDongle::getTablePrefix() . $this->related->getQualifiedKeyName()
```

src/Foundation/Providers/ExecutionContextProvider.php
```diff
@@ -34,14 +33,10 @@ public function register()
...
-            $url = '/'.$url;
-        }
-
         if (!strlen($url)) {
-            $url = '/';
+            return '/';
         }
 
-        return $url;
+        return (substr($url, 0, 1) !== '/') ? '/' . $url : $url;
```
LukeTowers: Should this be Str::startsWith() instead for better readability?

src/Translation/Translator.php
```diff
@@ -167,9 +167,13 @@ public function parseKey($key)
...
     {
-        $locales = array_values(array_filter([$locale ?: $this->locale, $this->fallback, static::CORE_LOCALE]));
+        $locales = array_values(parent::localeArray($locale));
+
+        if (!in_array(static::CORE_LOCALE, $locales)) {
+            $locales[] = static::CORE_LOCALE;
+        }
 
-        return call_user_func($this->determineLocalesUsing ?: fn () => $locales, $locales);
+        return $locales;
```
LukeTowers, line 176: Does this mean that the determineLocalesUsing feature doesn't work in Winter?

src/Database/Model.php
bennothommo: It's unreachable code.

src/Translation/Translator.php
bennothommo: It's called in the parent class. We're just appending our core language to the end if need be.

LukeTowers (approved):

PR #71, opened 2022-02-28
Fix Auth setPermissionsAttribute method having the wrong type-hint.

ericp-mrel:
The User/Role Auth model `permissions` attribute mutator (setPermissionsAttribute) contains the wrong type-hint in the doc block.

`json_decode` accepts a string as the first parameter, not an array.


LukeTowers:
Thanks @ericp-mrel!

PR #72, opened 2022-03-03
Revise class loader

LukeTowers:
Fixes https://github.com/wintercms/winter/discussions/214.

This fixes a long standing issue with the ClassLoader that handles loading module and plugin classes where only lowercase paths to the (properly cased) class file were technically supported. This prevents PSR-4 folder structures in plugins and modules from working; and even worse it would usually only show up in production with an opaque error message since production machines typically have case sensitive filesystems and development machines tend to not.

The current logic uses a directory scanning approach of generating many different possible variations of a valid path for a given class and then looping over the "loaded directories" to scan each of them for a valid instance of a possible path. 

The new logic instead switches to requiring class prefix to class path to be registered with the ClassLoader explicitly which not only makes the resolution logic much faster (since the prefix is directly matched with the request class instead of always having to scan for a potential match) but it also means that the autoloading can be completely disabled for plugins and modules that aren't currently enabled which isn't currently the case and will help with fully cutting off disabled plugins and modules from the application.


src/Support/ClassLoader.php
```diff
@@ -135,28 +114,35 @@ class_alias($class, $reverse);
             return true;
         }
 
-        list($lowerClass, $upperClass, $lowerClassStudlyFile, $upperClassStudlyFile) = static::getPathsForClass($class);
```
LukeTowers, line 138: Need to remove this method

src/Support/ClassLoader.php
```diff
@@ -135,28 +114,35 @@ class_alias($class, $reverse);
...
+
+                    // Fallback to the unmodified path
+                    $path . $directory . DIRECTORY_SEPARATOR . $file,
+                ];
+
+                foreach ($pathsToTry as $classPath) {
+                    if ($this->isRealFilePath($classPath)) {
+                        $this->includeClass($class, $classPath);
+
+                        if (!is_null($reverse = $this->getReverseAlias($class))) {
```
bennothommo: I'd prefer we define variables outside of conditional statements, so they can be analysed by Xdebug before going into the condition.
```suggestion
						$reverse = $this->getReverseAlias($class);
                        if (!is_null($reverse)) {
```

src/Support/ClassLoader.php
```diff
@@ -135,28 +114,35 @@ class_alias($class, $reverse);
...
-            foreach ($paths as $path) {
-                if ($this->isRealFilePath($path)) {
-                    $this->includeClass($class, $path);
-
-                    if (!is_null($reverse = $this->getReverseAlias($class))) {
-                        if (!class_exists($reverse, false) && !in_array($reverse, $this->loadedAliases)) {
-                            class_alias($class, $reverse);
-                            $this->reversedClasses[] = $reverse;
+        // Check our registered autoload packages for a match
+        foreach ($this->autoloadedPackages as $prefix => $path) {
```
LukeTowers, line 114: The autoloaded packages should probably be sorted by the number of `\` characters in the prefix to make sure more specific prefixes are looked at first.

damsfx:
@LukeTowers Is this PR responsible for the class loading issues in Windows?

LukeTowers:
@damsfx most likely yes. We still haven't encountered the issues you were reporting nor have we heard from anyone else about them so if you're able to track it down on your machine and submit a PR I would appreciate that :)

damsfx:
@LukeTowers 
After investigating ... the [resolvePath](https://github.com/wintercms/storm/blob/16b898b334aba8bec9581f515a75831d5c9bc3fc/src/Support/ClassLoader.php#L159) of the ClassLoader seems to be the faulty.

When running the Winter.Test plugin migration, I've got this issue.  
```bash 
In seed_tables.php line 31:
  Class "Winter\Test\Models\Person" not found
```

The `resolvePath()` method is called with the following argument in this case 
```php
$path = "P:\_Sites\_Labs\winterdev\plugins\winter\test\Models\Person.php"
```
but the return value is 
```php
$path = "P:\_Sites\_Labs\winterdev\P:\_Sites\_Labs\winterdev\plugins\winter\test\Models\Person.php"
```
The method adds `$this->basePath . DIRECTORY_SEPARATOR` unnecessarily.

It may be necesary to include the base path in the test : 
```php
    /**
     * Resolve the provided path, relative or absolute
     */
    protected function resolvePath(string $path): string
    {
        if (!Str::startsWith($path, ['/', '\\', $this->basePath . DIRECTORY_SEPARATOR])) {
            $path = $this->basePath . DIRECTORY_SEPARATOR . $path;
        }
        return $path;
    }
```

mjauvin:
Oh, I see, on windows the basePath does not necessarily start with a slash or backslash...

Care to submit a PR @damsfx ?

PR #73, opened 2022-03-04
Add Array Source trait

bennothommo:
This PR introduces the ability to create an "array source" model which provides arbitrary records via an array. These models will act, for all intents and purposes, as Eloquent models, but do not require a database.

The implementation of this feature is inspired by the [Sushi library by Caleb Porzio](https://github.com/calebporzio/sushi) but modified to work in Winter's architecture.

This PR also introduces a change to the testing suite, to cater for the facades used in this trait, and now creates a dummy application when running tests, which resolves a long-standing issue where facades in the test cases did not work.

TODO:
- [ ] Document change
- [x] Add unit tests for the new feature


composer.json
```diff
@@ -32,7 +32,8 @@
         "ext-libxml": "*",
         "ext-mbstring": "*",
         "ext-openssl": "*",
-        "ext-PDO": "*",
+        "ext-pdo": "*",
+        "ext-pdo_sqlite": "*",
```
LukeTowers: This shouldn't be added here, instead the relevant recommend package entry should be updated to include this feature as requiring this extension.

src/Database/Traits/ArraySource.php
```diff
@@ -0,0 +1,307 @@
...
+
+    /**
+     * Sets the array chunk size when storing inserts.
+     *
+     * Sometimes, SQLite will complain if given too many records to insert at once, so we will split the records up
+     * into reasonable chunks and insert them in groups.
+     *
+     * @return int
+     */
+    protected function getArrayChunkSize()
```
LukeTowers: Trait method names should be prefixed with the namespacing, so `arraySourceGetChunkSize()`, `arraySourceDbNeedsUpdate()`, etc.

tests/Database/Traits/ValidationTest.php
```diff
@@ -20,13 +23,16 @@ protected function getConnectionName()
...
-    protected function getTable()
+    protected function getTable($table = null)
     {
         return 'users';
     }
 
     public function testUnique()
     {
+        // Fake a request so flash messages are not sent
+        Input::swap(new Request());
```
LukeTowers, line 34: Isn't the Input facade just a wrapper around the Request facade? We should avoid using it at all, it exists purely as a BC layer if I recall correctly

tests/Halcyon/HalcyonModelTest.php
```diff
@@ -17,6 +19,9 @@ public function setUp(): void
 
         $this->setDatasourceResolver();
 
+        // Fake a request so flash messages are not sent
+        Input::swap(new Request());
```
LukeTowers, line 23: Another usage of Input

tests/Support/MailFakeTest.php
```diff
@@ -1,21 +1,16 @@
...
-            class_alias('\Illuminate\Support\Facades\App', 'App');
-        }
-        App::shouldReceive('getLocale')->andreturn('en/US');
+        parent::setUp();
 
-        // Mock Mail facade
-        if (!class_exists('Mail')) {
-            class_alias('\Winter\Storm\Support\Facades\Mail', 'Mail');
-        }
+        App::shouldReceive('getLocale')->andreturn('en/US');
```
LukeTowers: is `andreturn` the actual method or is it `andReturn`?

tests/TestCase.php
```diff
@@ -1,16 +1,27 @@
...
-     * @return Symfony\Component\HttpKernel\HttpKernelInterface
+     * @return \Winter\Storm\Foundation\Application
      */
-    public function createApplication()
+    protected function resolveApplication()
     {
+        return tap(new Application($this->getBasePath()), function ($app) {
+            $app->bind(
+                'Winter\Storm\Foundation\Bootstrap\LoadConfiguration',
+                'Orchestra\Testbench\Bootstrap\LoadConfiguration'
```
LukeTowers: For class strings, should we use the `::class` constant instead for better IDE integration?

composer.json
bennothommo: Sorry, I don't follow?

tests/Database/Traits/ValidationTest.php
bennothommo: It is being used in the validation traits that these tests (and the ones below) are for, so we'd have to change them there.

composer.json
LukeTowers: Adding pdo sqlite as a required php extension breaks the library for installs without that extension. Instead, we should just add to the description that we already have for this dependencies entry in the recommended dependencies package.

bennothommo: Given this will be core functionality though, won't it actually be a dependency going forward? We couldn't just leave it as a suggestion.

src/Database/Traits/ArraySource.php
mjauvin: to avoid name clashing, yeah, good point.

composer.json
LukeTowers: Unless a high traffic part of the backend or cms uses it then no. Technically speaking supporting sqlite is already core functionality, and look at how the assetic library handles this. If you want to enable this specific functionality then you need to provide the related dependency.

src/Database/Traits/ArraySource.php
bennothommo: Alright, but just to note, none of our other traits follow that yet.

tests/Support/MailFakeTest.php
bennothommo: Whoops, will fix

mjauvin:
Not sure why these tests are failing

LukeTowers:
@bennothommo will need to delete the branch on github and re-push up his local before merging 1.2 in properly.

PR #74, opened 2022-03-04
Use extension if provided in resizer options

mjauvin:
This is required to allow converting images to other formats (e.g. webp). Otherwise the resulting file will have the correct extension but the wrong image format.

Replaces #9 since the original repo got deleted.

The following code can be used as a workaround in the meantime:
```
Event::listen('system.resizer.processResize', function ($resizer, $tempPath) {
    // Get the resizing configuration
    $config = $resizer->getConfig();
    $options = array_get($config, 'options', []);

    list($base, $ext) = explode('.', $tempPath);
    $newPath = $base . '.' . array_get($options, 'extension', $ext);

    // Resize the image
    $resizedImageContents = \Winter\Storm\Database\Attach\Resizer::open($tempPath)
        ->resize($config['width'], $config['height'], $options)
        ->save($newPath);

    if ($newPath != $tempPath) {
        File::move($newPath, $tempPath);
    }

    // Prevent any other resizing replacer logic from running
    return true;
});
```

### DIFF
--- a/src/Database/Attach/Resizer.php
+++ b/src/Database/Attach/Resizer.php
@@ -693,7 +693,7 @@ class Resizer
      * @param string $path The filename path
      * @return string
      */
-    protected function getExtension($path)
+    protected function getExtension(string $path): string
     {
         return $this->getOption('extension') ?: (pathinfo($path, PATHINFO_EXTENSION) ?: $this->extension);
     }
--- a/src/Database/Attach/Resizer.php
+++ b/src/Database/Attach/Resizer.php
@@ -451,7 +451,7 @@ class Resizer
         }
 
         // Determine the image type from the destination file
-        $extension = pathinfo($savePath, PATHINFO_EXTENSION) ?: $this->extension;
+        $extension = $this->getExtension($savePath);
 
         // Create and save an image based on it's extension
         switch (strtolower($extension)) {
@@ -686,5 +686,15 @@ class Resizer
         $optimalHeight = round($this->height * $effectiveRatio);
 
         return [$optimalWidth, $optimalHeight];
+    }
+
+    /**
+     * Return the filename extension for the provided filename
+     * @param string $path The filename path
+     * @return string
+     */
+    protected function getExtension($path)
+    {
+        return $this->getOption('extension') ?: (pathinfo($path, PATHINFO_EXTENSION) ?: $this->extension);
     }
 }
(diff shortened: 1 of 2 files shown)


mjauvin:
@LukeTowers any reason against merging this? Converting a jpg to a webp won't work without this fix.

tests/Database/Attach/ResizerTest.php
```diff
@@ -72,6 +72,30 @@ public function testReset()
         $this->assertImageSameAsFixture(self::COMMON_FIXTURES['reset']);
     }
 
+    /**
+     * Given a Resizer with any image
+     * Test to see if the getExtension() method properly returns the filename extension for the provided filename
+     * @throws Exception
```
LukeTowers: This doesn't seem like it throws an exception does it?

src/Database/Attach/Resizer.php
```diff
@@ -687,4 +687,14 @@ protected function getSizeByFit($maxWidth, $maxHeight)
 
         return [$optimalWidth, $optimalHeight];
     }
+
+    /**
+     * Return the filename extension for the provided filename
+     * @param string $path The filename path
+     * @return string
+     */
+    protected function getExtension($path)
```
LukeTowers: Can you add the type hints to the method signature instead of the docblock?

tests/Database/Attach/ResizerTest.php
mjauvin: @LukeTowers This is what all TestCase test methods throw, isn't it?

LukeTowers: not if they're written correctly? Generally if a test method expects to throw an exception, it will call $this->expectException(); (or whatever the actual name is).

src/Database/Attach/Resizer.php
```diff
@@ -687,4 +687,14 @@ protected function getSizeByFit($maxWidth, $maxHeight)
 
         return [$optimalWidth, $optimalHeight];
     }
+
+    /**
+     * Return the filename extension for the provided filename
+     * @param string $path The filename path
+     * @return string
```
LukeTowers: We don't want to keep docblock typehints around if we have method signature typehints, makes it harder to maintain.

mjauvin:
@LukeTowers why did you close this?

bennothommo:
@mjauvin I think these were automatically closed when Luke deleted the `wip/1.2` branch. Would you be able to recreate the PRs targeting `develop`?

LukeTowers:
@bennothommo is correct, it was unintentionally auto closed.

PR #80, opened 2022-04-07
Restored mail dispatcher being passed correct message obj

jaxwilko:
PR fixes the following error `Call to undefined method Illuminate\Mail\SentMessage::getSymfonyMessage()` when attempting to send mail via the reset password feature.

The issue is we're calling `dispatchSentEvent()` in `src/Mail/Mailer.php:161` and passing `Illuminate\Mail\SentMessage` instead of `Illuminate\Mail\Message`.


jaxwilko:
Closing this as it's not needed since https://github.com/laravel/framework/commit/63ca843643e86fb69efc901051ae079c89a7fd09

PR #81, opened 2022-04-12
Can't pass null to non-nullable arguments

Denoder:
extension of: https://github.com/wintercms/winter/pull/524

There should be a better way of handling this.


src/Support/helpers.php
```diff
@@ -20,7 +20,7 @@ function e($value, $doubleEncode = false)
             return $value->toHtml();
         }
 
-        return htmlspecialchars($value, ENT_QUOTES, 'UTF-8', $doubleEncode);
+        return htmlspecialchars($value ?? '', ENT_QUOTES, 'UTF-8', $doubleEncode);
```
LukeTowers:
```suggestion
        return htmlspecialchars((string) $value, ENT_QUOTES, 'UTF-8', $doubleEncode);
```

PR #86, opened 2022-06-03
Load deferred providers before provider registration

bennothommo:
Implements https://github.com/laravel/framework/pull/38724 into the Winter codebase, extending the base Laravel ProviderRepository class and overriding the `load` method. Reverts https://github.com/wintercms/storm/commit/dff4e6e2f150c059fd1934e4cac3dcbc99af638e and partially reverts https://github.com/wintercms/storm/commit/6d950462bf8aaf8a3d5189030797b6b39b2ea7d6.

This PR will override the base provider loading functionality to record deferred providers *before* registering any eager loaded providers. This allows eager loaded providers to use deferred provider functionality (such as the Cache provider). In this method, this will technically make those providers eager loaded as well, but these providers will be deferred if no other provider uses them in registration.


LukeTowers (approved):
@bennothommo if you've tested with the latest 1.2 of the system module and it works fine in the web environment I'm happy to have this merged, seems like a pretty great solution!

PR #88, opened 2022-06-15
Add unit test for overriden to() method

mjauvin:
ref: https://github.com/wintercms/wn-translate-plugin/issues/36#issuecomment-1156430753


arvislacis:
Although added test looks valid and useful, the actual result of failed cases is the same, I mean, if you decode `https://www.example.com/path?array%5B%5D=v1&array%5B%5D=v2` then it's equal to `https://www.example.com/path?array[]=v1&array[]=v2`

Problems starts with `https://www.example.com/path?array[0]=v1&array[1]=v2` which is then interpretated as `https://www.example.com/path?array[0][]=v1&array[1][]=v2`

mjauvin:
@arvislacis Yes, I am aware of this, but @LukeTowers needed a test case that showed the issue in order to fix the problem.

LukeTowers:
Merged in https://github.com/wintercms/storm/commit/32df10c002b7b522bdac054fc991d74b73533f98

PR #89, opened 2022-06-17
Added support for setting system local temp path via config

jaxwilko:
This PR allows the dev to set the application temp path (`temp_path()`) to be set via `filesystems.localTempPath` config.

### DIFF
--- a/src/Foundation/Bootstrap/RegisterWinter.php
+++ b/src/Foundation/Bootstrap/RegisterWinter.php
@@ -37,7 +37,7 @@ class RegisterWinter
             $app->setThemesPath($themesPath);
         }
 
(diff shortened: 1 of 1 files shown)


src/Foundation/Bootstrap/RegisterWinter.php
```diff
@@ -36,5 +36,9 @@ public function bootstrap(Application $app)
         if ($themesPath = $app['config']->get('cms.themesPathLocal')) {
             $app->setThemesPath($themesPath);
         }
+
+        if ($tempPath = $app['config']->get('filesystems.localTempPath')) {
```
LukeTowers:
```suggestion
        if ($tempPath = $app['config']->get('app.tempPath')) {
```

PR #91, opened 2022-06-27
Bring Translator closer to base Laravel functionality, add "set" method

bennothommo:
Refactors the "get" override to use the base "get" method for the most part, removes the "choice" method override and adds the "set" method in Translator.

Changes to the FileLoader will allow our original fix for the sublocales in the "choice" method to work even with the base Laravel functionality.

The FileLoader will now attempt to find locales in the format "xx_XX" as per the original Laravel functionality and then also attempt "xx-xx" as per the traditional usage in Winter.

Fixes https://github.com/wintercms/winter/issues/480

### DIFF
--- a/src/Translation/FileLoader.php
+++ b/src/Translation/FileLoader.php
@@ -29,10 +29,12 @@ class FileLoader extends FileLoaderBase
         // Try "xx-xx" format
         $locale = str_replace('_', '-', strtolower($locale));
 
-        $file = "{$this->path}/{$locale}/{$namespace}/{$group}.php";
+        if ("{$this->path}/{$locale}/{$namespace}/{$group}.php" !== $file) {
+            $file = "{$this->path}/{$locale}/{$namespace}/{$group}.php";
 
-        if ($this->files->exists($file)) {
-            return array_replace_recursive($lines, $this->files->getRequire($file));
+            if ($this->files->exists($file)) {
+                return array_replace_recursive($lines, $this->files->getRequire($file));
+            }
         }
 
         return $lines;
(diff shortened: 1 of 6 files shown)


LukeTowers:
Can we pull in the unit test from Laravel for the translator? Also were you able to test what happens if a key is set before it's namespace is loaded? In raw Laravel 6 I found that would cause the loader to think the entire namespace had been loaded before it actually was leading to some convoluted workarounds that it would be nice to avoid in ours.

```php
if (!empty($project->localizationMessages)) {

                // loop over locales
                foreach ($project->localizationMessages as $locale => $values) {
                    $groups = [];
                    $messages = [];
                    // loop over messages, adding groups
                    foreach ($values as $k => $v) {
                        $group = Str::before($k, '.');
                        $messages[$k] = $v;
                        if (!in_array($group, $groups)) {
                            $groups[] = $group;
                        }
                    }
                    // Ensure each group that will be modified is loaded to avoid issues with the Translator thinking
                    // the overrides are the entire scope of the group and not loading the rest of the group.
                    foreach ($groups as $group) {
                        Lang::load('*', $group, $locale);
                    }
                    // add message overrides
                    Lang::addLines($messages, $locale);
                }
            }       
```

bennothommo:
@LukeTowers unless I'm mistaken, the `set` method is writing to where the JSON translation definition is stored, which operates independently from the PHP translation files, and wouldn't require us to load the namespaced files. The `get` method checks this first, then goes down to the PHP files if it doesn't find a translation there.

As long as we load up the JSON translation file first, we should be right. I will create a unit test to check this though.

bennothommo:
@LukeTowers these two test cases test your scenario, and both are working:
https://github.com/wintercms/storm/pull/91/files#diff-816079b94385b284b03f5bc82e497d79d0cba0ebad5628a858da771861e92f65R160-R185

LukeTowers:
@bennothommo can we get rid of the translator.beforeResolve event now?

LukeTowers (approved):

PR #92, opened 2022-07-04
Restore Laravel 6 behavior to public visibility for local driver

mjauvin:
Fixes https://github.com/wintercms/winter/issues/503

@LukeTowers @bennothommo @jaxwilko

Here's my preferred solution to this.

### DIFF
--- a/src/Filesystem/FilesystemManager.php
+++ b/src/Filesystem/FilesystemManager.php
@@ -28,7 +28,7 @@ class FilesystemManager extends BaseFilesystemManager
         if (is_null($config)) {
             $config = $this->getConfig($name);
         }
-        if ($name === 'local') {
+        if ($config['driver'] === 'local' && empty($config['visibility']))
             $config['visibility'] = 'public';
         }
 
--- a/src/Filesystem/FilesystemManager.php
+++ b/src/Filesystem/FilesystemManager.php
@@ -22,4 +22,16 @@ class FilesystemManager extends BaseFilesystemManager
         }
         return $configName;
     }
+
+    protected function resolve($name, $config = null)
+    {
+        if (is_null($config)) {
+            $config = $this->getConfig($name);
+        }
+        if ($name === 'local') {
+            $config['visibility'] = 'public';
+        }
+
+        return parent::resolve($name, $config);
+    }
 }
--- a/src/Filesystem/FilesystemManager.php
+++ b/src/Filesystem/FilesystemManager.php
@@ -31,6 +31,9 @@ class FilesystemManager extends BaseFilesystemManager
         if (is_null($config)) {
             $config = $this->getConfig($name);
         }
+
+        // Default local drivers to public visibility for backwards compatibility
+        // see https://github.com/wintercms/winter/issues/503
         if ($config['driver'] === 'local' && empty($config['visibility'])) {
             $config['visibility'] = 'public';
         }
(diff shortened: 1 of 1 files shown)


src/Filesystem/FilesystemManager.php
```diff
@@ -22,4 +22,16 @@ public function identify($disk)
         }
         return $configName;
     }
+
+    protected function resolve($name, $config = null)
+    {
+        if (is_null($config)) {
+            $config = $this->getConfig($name);
+        }
+        if ($name === 'local') {
```
LukeTowers: Shouldn't this be the following instead?
```php
if ($config['driver'] === 'local' && empty($config['visibility']))
```

LukeTowers:
@driesvints if you have a moment do you recall or have a link to the reasoning behind changing the default from public to private for the local Flysystem driver in Laravel 9?

driesvints:
@LukeTowers that got changed in Flysystem v3 so you'll have to ask them. Afaik it's a more safe and sensible default. You should always use the same user to handle files.

LukeTowers:
@driesvints I agree in theory that it's more safe default, the PHP user can access the files but the problem comes when the webserver user wants to fetch them to respond to requests, which is the default used by a few providers, ServerPilot & Cleavr for example. 

Any recommendations for how to handle that use case?

driesvints:
Not really sorry..

src/Filesystem/FilesystemManager.php
mjauvin: That works too

LukeTowers:
Thanks anyways :)

src/Filesystem/FilesystemManager.php
LukeTowers:
```suggestion
        if ($config['driver'] === 'local' && empty($config['visibility']))
```

src/Filesystem/FilesystemManager.php
```diff
@@ -22,4 +22,16 @@ public function identify($disk)
         }
         return $configName;
     }
+
```
LukeTowers, line 25:
```suggestion

    /**
     * @inheritDoc
     */
```

src/Filesystem/FilesystemManager.php
```diff
@@ -22,4 +22,19 @@ public function identify($disk)
...
+    /**
+     * @inheritDoc
+     */
+    protected function resolve($name, $config = null)
+    {
+        if (is_null($config)) {
+            $config = $this->getConfig($name);
+        }
+        if ($config['driver'] === 'local' && empty($config['visibility']))
```
LukeTowers:
```suggestion
        if ($config['driver'] === 'local' && empty($config['visibility'])) {
```

src/Filesystem/FilesystemManager.php
```diff
@@ -22,4 +22,19 @@ public function identify($disk)
...
+
+    /**
+     * @inheritDoc
+     */
+    protected function resolve($name, $config = null)
+    {
+        if (is_null($config)) {
+            $config = $this->getConfig($name);
+        }
+        if ($config['driver'] === 'local' && empty($config['visibility'])) {
```
LukeTowers, line 37:
```suggestion

        // Default local drivers to public visibility for backwards compatibility
        // see https://github.com/wintercms/winter/issues/503
        if ($config['driver'] === 'local' && empty($config['visibility'])) {
```

PR #93, opened 2022-07-05
[WIP] 1.2 - Fix Twig Compiling

jaxwilko:
The PR adds a hack that fixes an exception thrown during twig rendering.

> This is not the proper fix, the proper fix is: https://github.com/twigphp/Twig/pull/3719

This should be removed once @bennothommo's PR is merged.

### DIFF
--- a/src/Foundation/Application.php
+++ b/src/Foundation/Application.php
@@ -570,4 +570,15 @@ class Application extends ApplicationBase
          */
         return '';
     }
+
+    /**
+     * This is a temporary fix for an issue with twig reflection.
+     * The full fix is here: https://github.com/twigphp/Twig/pull/3719
+     *
+     * @TODO: Remove this after Twig PR 3719 is merged.
+     */
+    public function __toString(): string
+    {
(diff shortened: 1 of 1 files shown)


jaxwilko:
@mjauvin could you see if this fixes the issue for you? It's only a temporary solution but it's better than nothing :)

bennothommo:
@LukeTowers @jaxwilko I personally would prefer we apply the fix I suggested in the original issue (https://github.com/wintercms/winter/issues/557#issuecomment-1159670223) and use a call to a concrete class for the callback. I don't believe Twig will apply our fix anytime soon because it's such an edge case.

LukeTowers:
@bennothommo while I'd somewhat prefer using concrete classes over facades in general to some extent that isn't a fully workable solution for all our users given that they could very easily have been using their own functions defined from facades, so the underlying issue would still not be fixed.

bennothommo:
Fair call. It's a moot point now anyway, as it looks like it's being fixed at the PHP lang level.

PR #97, opened 2022-07-14
Model: make string attribute value auto-trimming optional

lex0r:
because not all attributes are made equal.

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -50,9 +50,9 @@ class Model extends EloquentModel
     public $duplicateCache = true;
 
     /**
-     * @var bool Indicates if all model attributes will be trimmed prior to saving.
+     * @var bool Indicates if all string model attributes will be trimmed prior to saving.
      */
-    public $trimAttributes = true;
+    public $trimStringAttributes = true;
 
     /**
      * @var array The array of models booted events.
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -50,6 +50,11 @@ class Model extends EloquentModel
     public $duplicateCache = true;
 
     /**
+     * @var bool Indicates if all model attributes will be trimmed prior to saving.
+     */
+    public $trimAttributes = true;
+
+    /**
      * @var array The array of models booted events.
      */
     protected static $eventsBooted = [];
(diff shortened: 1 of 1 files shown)


LukeTowers:
@lex0r can you add the test case from the original PR? https://github.com/octobercms/library/pull/503/files#diff-270f442232291e973d322f9166dda6fb544ad2325f6db11581f7d553171213a8

Also I'm curious to know your use case and what problems auto trimming was causing for you, are you able to share some details?

src/Database/Model.php
```diff
@@ -49,6 +49,11 @@ class Model extends EloquentModel
      */
     public $duplicateCache = true;
 
+    /**
+     * @var bool Indicates if all model attributes will be trimmed prior to saving.
+     */
+    public $trimAttributes = true;
```
```suggestion
    /**
     * @var bool Indicates if all string model attributes will be trimmed prior to saving.
     */
    public $trimStringAttributes = true;
```

src/Database/Model.php
```diff
@@ -1239,7 +1244,7 @@ public function setAttribute($key, $value)
         /*
          * Trim strings
          */
-        if (is_string($value)) {
+        if ($this->trimAttributes && is_string($value)) {
```
```suggestion
        if ($this->trimStringAttributes && is_string($value)) {
```

RomainMazB:
> @lex0r can you add the test case from the original PR? https://github.com/octobercms/library/pull/503/files#diff-270f442232291e973d322f9166dda6fb544ad2325f6db11581f7d553171213a8
> 
> Also I'm curious to know your use case and what problems auto trimming was causing for you, are you able to share some details?

Great addition here!!

As I remember, my use case was forum messages including markdown: if you start a message with a code block using tab instead of three backticks, the tabs was trimmed and the code block just didn't work...

lex0r:
> @lex0r can you add the test case from the original PR? https://github.com/octobercms/library/pull/503/files#diff-270f442232291e973d322f9166dda6fb544ad2325f6db11581f7d553171213a8
> 
> Also I'm curious to know your use case and what problems auto trimming was causing for you, are you able to share some details?

@LukeTowers I had to save a file to a remote location directly without any pre-processing and then save a record linking to it in the database. The file name can contain trailing spaces and that is acceptable, however when trying to save such a name in the database you would get the name trimmed. Later on we used the name to construct a link to the file and potential spaces were not included in that link which made it impossible to locate the file in the remote storage which has the file saved with the space.

LukeTowers:
Interesting, thanks for sharing your use case @lex0r!

PR #98, opened 2022-07-20
Added a __toString to resolve Twig issue

damsfx:
wintercms/storm@1b866f0  backported to 1.1 branch

### DIFF
--- a/src/Foundation/Application.php
+++ b/src/Foundation/Application.php
@@ -543,4 +543,15 @@ class Application extends ApplicationBase
     {
         return PathResolver::join($this->storagePath(), '/framework/classes.php');
     }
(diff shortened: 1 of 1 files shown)


LukeTowers:
@bennothommo or @jaxwilko are you able to fix the tests? They're failing on develop too so something must have changed in the underlying dependencies.

LukeTowers:
@jaxwilko ping

bennothommo:
@LukeTowers I think it should be fine - it seems to be only the image tests that failed. I can fix that on the 1.1 branch if need be. Seems to be fine on Windows too, so it's probably a subtle difference with one of the image codecs.

PR #103, opened 2022-08-01
Add is_regex validator rule

mjauvin:
credit @jaxwilko for this nice implementation.

used in `wn-redirect-plugin` to validate the regex in the Redirect model.
ref. https://github.com/wintercms/wn-redirect-plugin/blob/main/models/Redirect.php#L179


mjauvin:
Close as this is added as a custom validator from within the wn-redirect-plugin.

PR #104, opened 2022-08-11
Re-do image resizer tests

bennothommo:
Fixes up some issues with the Image Resizer tests. I intend to also port this to the develop branch for 1.2 as well, but 1.1 is the one currently failing.

- Replaces fixtures with Winter logo
- Increases the sizes tested for most cases. This should result in less failures when the underlying image codecs change, as what occurs between distribution updates or different OS's.
- Update validation fixtures based on test results
- Remove a test that relied on EXIF and is usually skipped


bennothommo:
Moving back to draft because apparently Windows doesn't like this.

PR #105, opened 2022-08-17
Add withoutChildren scope to NestedTree

RomainMazB:
Currently building a `Winter.User` blogs plugin, based on `Winter.Blog`.
I needed to list all the available parent categories for the current one, but excluding the current category's children to avoid circular references.

I thought it would be good to add this scope to the core.

### DIFF
--- a/src/Database/Traits/NestedTree.php
+++ b/src/Database/Traits/NestedTree.php
@@ -440,6 +440,22 @@ trait NestedTree
 
         return $includeSelf ? $query : $query->withoutSelf();
     }
+    
+    /**
+     * Extracts all children & nested children from current query expression.
+     *
+     * @param \Illuminate\Database\Query\Builder $query
+     * @return \Illuminate\Database\Query\Builder
+     */
+    public function scopeWithoutChildren($query, $includeSelf = false)
+    {
+        $query->whereNot([
+            [$this->getLeftColumnName(), '>', $this->getLeft()],
+            [$this->getRightColumnName(), '<', $this->getRight()]
+        ]);
+
+        return $includeSelf ? $query : $query->withoutSelf();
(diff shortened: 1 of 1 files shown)


LukeTowers:
@RomainMazB can you add a unit test for this?

RomainMazB:
All the actual tests are included in [wintercms/winter](https://github.com/wintercms/winter/blob/develop/modules/system/tests/plugins/database/NestedTreeModelTest.php), is there where you want me to add it?

_As a side note, it could be more clear to pull back almost all of [this wintercms/winter test folder](https://github.com/wintercms/winter/tree/develop/modules/system/tests/plugins/database) inside winter/storm as most of the tests are more related to storm than winter IMO_

bennothommo:
@RomainMazB

> As a side note, it could be more clear to pull back almost all of [this wintercms/winter test folder](https://github.com/wintercms/winter/tree/develop/modules/system/tests/plugins/database) inside winter/storm as most of the tests are more related to storm than winter IMO

I agree with this. It's something on my list to do, but the list is loooooooong, so won't be any time soon.

LukeTowers:
@RomainMazB the tests have been reorganized in the Winter core now, are you able to add a test case for this? Ideally at some point we should probably be able to run database tests on this library directly but for now a PR to the wintercms/winter with the test case for this addition and perhaps an addition to the docs and we should be able to merge this.

bennothommo:
Just to note, we can run database tests in Storm. There's a couple of them in there now - https://github.com/wintercms/storm/tree/develop/tests/Database/Traits. They just need to extend the `DbTestCase` base test class.

LukeTowers:
@RomainMazB are you still interested in getting this merged?

PR #107, opened 2022-08-25
Reverse order in checking mail config styles

### LINKED ISSUE #626: Winter 1.2 not using SMTP settings from database
### Winter CMS Build

1.2

### PHP Version

8.0

### Database engine

MySQL/MariaDB

### Plugins installed

Winter.Pages, Winter.Blog, Winter.User

### Issue description

Since upgrading to Winter 1.2 from the previous version, emailing via SMTP has not worked.

The error message when emailing has always been:
Expected response code "250/251/252" but got code "550", with message "550 5.7.1 Relaying denied".

I've looked into why and I have found that the email settings entered in the Winter backend, saved to the database, are not being used when actually emailing within Winter.

This can be tested by pressing the "Send test message" button in the "Mail configuration" backend section.

Upon looking at the code, when "createSmtpTransport($config)" in MailManager.php from laravel is called, the values it is looking for are $config['host'], $config['username'], $config['password'], $config['port'].

Ref: https://github.com/laravel/framework/blob/8156743f644fb4597b63261a3fda4e4d6ca49a78/src/Illuminate/Mail/MailManager.php#L163-L183

However, when "applyConfigValues()" in MailSetting.php is called, both the values from config/mail.php and the database are combined.

Ref: https://github.com/wintercms/winter/blob/develop/modules/system/models/MailSetting.php#L86-L118

The config values that get passed to laravel are in this format:
Array ( [driver] => smtp [host] => smtp.mailgun.org [port] => 587 [from] => Array ( [address] => mailaddress [name] => Winter CMS ) [encryption] => tls [username] => [password] => [sendmail] => /usr/sbin/sendmail -bs [default] => smtp [mailers] => Array ( [smtp] => Array ( [host] => smtp.office365.com [port] => 587 [username] => mailaddress [password] => password [encryption] => tls ) ) )

It seems that laravel then only uses the config values from config/mail.php instead of the database to make the Smtp connection.

### Steps to replicate

Leave the config/mail.php as default from the Winter installation.
Add Smtp settings into the Mail configuration section in the Winter CMS and click on "Send Test email".
Error message "Expected response code "250/251/252" but got code "550", with message "550 5.7.1 Relaying denied"." is then presented.

### Workaround

As a temporary workaround, put your Smtp configuration values into config/mail.php.

bennothommo:
Fixes https://github.com/wintercms/winter/issues/626

This reverses the order for checking mail configs by searching for the new format first, then checking for the old format.

Laravel checks old format first, then new format, which works fine for them, but doesn't work for us as we also populate settings via the Backend, which populates the settings in the new format.

There is a small gotcha here - I believe the Backend stored the config in the old format pre v1.2, which means if someone upgrades their `config/mail.php` to follow the new format, this will make the config file take precedence. Saving the settings in the Backend again will alleviate this.

### DIFF
--- a/src/Mail/MailManager.php
+++ b/src/Mail/MailManager.php
@@ -94,7 +94,7 @@ class MailManager extends BaseMailManager
     {
         // We will do the reverse of what Laravel does and check for "default" first, which is
         // populated by the Backend or the new "mail" config, before searching for the "driver"
-        // key that was present in older version of Winter.
+        // key that was present in older version of Winter (<1.2).
         return $this->app['config']['mail.default'] ?? $this->app['config']['mail.driver'];
     }
 }
--- a/src/Mail/MailManager.php
+++ b/src/Mail/MailManager.php
@@ -7,6 +7,7 @@ use Illuminate\Mail\MailManager as BaseMailManager;
  * Overrides the Laravel MailManager
  * - Replaces the Laravel Mailer class with the Winter Mailer class
  * - Fires mailer.beforeRegister & mailer.register events
+ * - Uses another method to determine old vs. new mail configs
  */
 class MailManager extends BaseMailManager
 {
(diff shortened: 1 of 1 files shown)


LukeTowers:
@bennothommo could you go into more detail on this part:
> There is a small gotcha here - I believe the Backend stored the config in the old format pre v1.2, which means if someone upgrades their config/mail.php to follow the new format, this will make the config file take precedence. Saving the settings in the Backend again will alleviate this.

bennothommo:
@LukeTowers never mind that comment - I've just checked and @mjauvin's PR to the [MailSetting](https://github.com/wintercms/winter/pull/522) model covers this scenario, so it should be fine to merge.

src/Mail/MailManager.php
```diff
@@ -71,4 +72,29 @@ protected function resolve($name)
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getDefaultDriver()
+    {
+        // We will do the reverse of what Laravel does and check for "default" first, which is
+        // populated by the Backend or the new "mail" config, before searching for the "driver"
+        // key that was present in older version of Winter.
```
LukeTowers:
```suggestion
        // key that was present in older version of Winter (<1.2).
```

PR #109, opened 2022-09-01
Added support for including anon classes multiple times per execution

jaxwilko:
`require_once` will fail to load anon classes on subsequent requests. This patch adds a caching layer that allows included classes and object to he held in memory and returned as needed

### DIFF
--- a/src/Database/Updater.php
+++ b/src/Database/Updater.php
@@ -13,6 +13,8 @@ use Winter\Storm\Database\Model;
  */
 class Updater
 {
+    protected static $migrationCache = [];
+
     /**
      * Sets up a migration or seed file.
      */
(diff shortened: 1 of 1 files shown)


bennothommo:
@jaxwilko I reckon we don't need to treat concrete classes any differently from anonymous classes. You could probably just create a cached instance of those classes as well, then there's no need to keep checking if it's an object or not when retrieving from the cache.

jaxwilko:
@bennothommo yeah, was just trying to reduce memory usage by holding class references rather than the objects where possible but I guess it shouldn't make a massive difference :)

LukeTowers:
@bennothommo I'd prefer not to, it's unnecessary memory usage to avoid having to initialize a new instance. I'm fine with this approach as it is.

src/Database/Updater.php
```diff
@@ -13,6 +13,8 @@
  */
 class Updater
 {
+    protected static $migrationCache = [];
```
LukeTowers, line 19:
```suggestion
    /**
     * @var array Local cache of migration file paths to support anonymous migrations [$path => $anonInstance || $className]
     */
    protected static $migrationCache = [];
```

LukeTowers (approved):

PR #110, opened 2022-09-02
Ensure cached routes are loaded on app boot

### LINKED ISSUE #666: Route cache broken on winter 1.2
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

Running `php artisan route:cache` breaks all routes.

HTTP requests return NotFoundHttpException and `php artisan route:list` returns "Your application doesn't have any routes."

### Steps to replicate

- Fresh install of wintercms
- Run `php artisan route:cache`

### Workaround

Clear the route cache by running `php artisan route:clear`

bennothommo:
Simplifies the routing service provider and checks if routes are cached. If so, this will load the cached routes on the app being booted.

Fixes https://github.com/wintercms/winter/issues/666


LukeTowers:
Do we not need the UrlGenerator override stuff in place for our functionality to take effect?

bennothommo:
Don't believe so :)

canthis:
Tested this one. It appears that routes provided by a plugin are not being cached (e.g. `vendor/plugin/routes.php`)

bennothommo:
@canthis I don't believe they've ever been cached, or picked up for that matter, since it's routed through the main Backend Controller class.

LukeTowers:
@bennothommo those are different, @canthis is talking about routes that are explicitly registered by the plugins' `routes.php` file.

bennothommo:
@LukeTowers @canthis happy to tackle that in another PR, this one should at least fix the immediate issue of the site breaking if someone uses `route:cache`.

LukeTowers:
@bennothommo I still don't understand why we need to remove the UrlGenerator override code, could you explain that?

bennothommo:
@LukeTowers I thought our extended `UrlGenerator` just adds the static method `buildUrl`, so it wouldn't matter, but I just realised that we also overwrite the `to` method. So I'll probably have to add that block back into the service provider.

LukeTowers:
@bennothommo are you still interested in finishing this PR?

ArturCiolkiewicz:
@bennothommo Thank You

This fixes problem with cached routes, but similar problem occurs with configuration that is not available from the cache.

php artisan config:cache or php artisan optimize
Winter 1.2.1 or dev-develop

bennothommo:
Thanks for letting us know @ArturCiolkiewicz - could you post that as a separate issue for us? I'll be able to sort it out within a couple of days.

PR #112, opened 2022-09-07
Add extendable.afterConstruct event

mjauvin:
The event fires once the behaviors have been loaded.

### DIFF
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -113,7 +113,9 @@ trait ExtendableTrait
          *     });
          *
          */
-        $this->fireEvent('extendable.afterConstruct');
+        if (method_exists($this, 'fireEvent')) {
+            $this->fireEvent('extendable.afterConstruct');
+        }
     }
 
     /**
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -101,6 +101,19 @@ trait ExtendableTrait
 
             $this->extendClassWith($useClass);
         }
+
+        /**
+         * @event extendable.afterConstruct
+         * Called after the dynamic behaviors have been loaded.
+         *
+         * Example usage:
+         *
+         *     $model->bindEvent('extendable.afterConstruct', function () use (\Winter\Storm\Database\Model $model) {
+         *         \Log::info(get_class($model) . ' has booted and behaviors are loaded');
+         *     });
+         *
+         */
+        $this->fireEvent('extendable.afterConstruct');
     }
 
     /**
--- a/tests/Extension/ExtendableTest.php
+++ b/tests/Extension/ExtendableTest.php
@@ -355,6 +355,12 @@ class ExtendableTestExampleExtendableClass extends Extendable
     {
         return $this->protectedFoo;
     }
+
+    // dummy method to appease test case on mock object
+    public function fireEvent($event)
+    {
+        return $event;
+    }
 }
 
 /*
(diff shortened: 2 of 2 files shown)


mjauvin:
@bennothommo can we add a `fireEvent()` method to the mockup or something?

src/Extension/ExtendableTrait.php
```diff
@@ -101,6 +101,19 @@ public function extendableConstruct()
...
+         * Called after the dynamic behaviors have been loaded.
+         *
+         * Example usage:
+         *
+         *     $model->bindEvent('extendable.afterConstruct', function () use (\Winter\Storm\Database\Model $model) {
+         *         \Log::info(get_class($model) . ' has booted and behaviors are loaded');
+         *     });
+         *
+         */
+        $this->fireEvent('extendable.afterConstruct');
```
LukeTowers: This needs to be wrapped in a method_exists check

mjauvin: fixed

LukeTowers:
@bennothommo @jaxwilko any input on this?

bennothommo:
@mjauvin @LukeTowers I just want to be sure that you are wanting this to fire on every single instance of an extendable class (not just models)?

LukeTowers:
@bennothommo yes, in my case that I needed it for I wanted to be able to call addAssets() on a widget implementing a behaviour to implement the required assets for that behaviour so long as it's conditions for enabling itself were met. However, I was unable to call the methods that behaviour added to the widget (which were required to determine if it was eligible to run and thus add the assets in the constructor of the behaviour. 

Perhaps we could also implement this by calling `init()` or `boot()` on the behavior after all the behaviours are registered instead of an event? @mjauvin would that work for your use case?

bennothommo:
@LukeTowers personally, I'd introduce another parameter to `extend` to specify that a callback should be run *after* the implements have been applied. That way, we're not doing an event on every single instance.

LukeTowers:
@bennothommo I was thinking more along the lines of Laravel's bootable Eloquent traits, i.e. the behaviour would add a method called `bootNameOfClass()` that gets called after all the behaviours have been registered (see https://andy-carter.com/blog/using-laravel-s-eloquent-traits). That way there's no added complexity to the API and there's no extra overhead of the events system.

bennothommo:
@LukeTowers that's a much more wild change to the current API than my suggestion though 😜

LukeTowers:
Is it? It's fairly unobtrusive IMO, it's an optional method that be implemented by the behaviour to hook into after all of the behaviours have been registered.

bennothommo:
How would you propose that someone leverages it? Would it be part of an `extend` callback, or another method altogether?

My suggestion is simply an extra parameter to `extend`, with a default that keeps BC. I'd argue it's more graceful.

LukeTowers:
I don't remember what @mjauvin's use case of it was, but it wasn't so much external usage of it but instead internal usage by the behaviour itself in order to do setup tasks that require the method proxying to be in place on the object that's implementing them before they can do certain tasks.

mjauvin:
My use case is to be able to call behavior's methods within the extend() callback. So @bennothommo's suggestion would work for me I believe.

mjauvin:
I'd like to be able to do the following:
```
class MyModel extends Model {
    public $implement = ['MyBehavior'];
}

MyModel::extend(function ($model) {
    $model->myBeaviorMethod();
})
```

mjauvin:
Replaced by #113

PR #113, opened 2022-09-08
Make extend() method usable with behaviors.

mjauvin:
Replaces #112 

```php
Model::extend(function ($model) {
    $model->behaviorMethodAvailable();
}, after: true);
```

Related: https://github.com/wintercms/docs/pull/98

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -147,7 +147,7 @@ class Model extends EloquentModel implements ModelInterface
     /**
      * Extend this object properties upon construction.
      */
-    public static function extend(Closure $callback, $after = false)
+    public static function extend(callable $callback, bool $after = false)
     {
         self::extendableExtendCallback($callback, $after);
     }
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -147,9 +147,9 @@ class Model extends EloquentModel implements ModelInterface
     /**
      * Extend this object properties upon construction.
      */
-    public static function extend(Closure $callback)
+    public static function extend(Closure $callback, $after = false)
     {
-        self::extendableExtendCallback($callback);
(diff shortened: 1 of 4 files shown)


bennothommo:
@mjauvin I've tweaked the property names and did some clean up of the Extendable trait. Feel free to revert if you disagree with it all :)

mjauvin:
So, who's in favor of merging this? Or something similar but through a seperate dedicated method ( e.g. `extendPostBehavior()` )

bennothommo (approved):

LukeTowers:
@mjauvin are you needing this in 1.2.1 or are you fine with waiting for 1.2.2?

mjauvin:
@LukeTowers I'm fine to wait for 1.2.2, which is why I moved it there.

mjauvin:
Should I create a PR to the docs repo documenting this?

bennothommo:
Just to note - this PR could be implemented via https://github.com/wintercms/storm/pull/134 instead if we wanted to go down that route.

@mjauvin's use case could be simply extending a particular instance  of the model locally via `$model->extend()`, which happens after instantiation and so other methods created by behaviors or other extensions would already be available.

mjauvin:
> Just to note - this PR could be implemented via #134 instead if we wanted to go down that route.
> 
> @mjauvin's use case could be simply extending a particular instance of the model locally via `$model->extend()`, which happens after instantiation and so other methods created by behaviors or other extensions would already be available.

@bennothommo are you saying I could do:
```
MyModel::extend(function () {
    $this->extend(function () {
        $this->behaviorMethodAvailable();
    });
}, true);
```

And that `$this->behaviorMethodAvailbale()` call would work?

mjauvin:
Close in favor of #134.

PR #115, opened 2022-09-18
Translation of messages with authorization errors

WebVPF:
Backend login not translated authorization error messages:

This commit makes the messages available for translation.

### DIFF
--- a/src/Auth/Manager.php
+++ b/src/Auth/Manager.php
@@ -6,6 +6,7 @@ use Illuminate\Support\Facades\Session;
 use Illuminate\Support\Facades\Request;
 use Illuminate\Contracts\Auth\Authenticatable;
 use Illuminate\Session\SessionManager;
+use Lang;
 
 /**
  * Authentication manager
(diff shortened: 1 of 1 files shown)


LukeTowers:
Thanks for your interest in contributing to Winter CMS!

This has been discussed a few times before but it boils down to this:
- Detailed authentication messages are not meant for general usage consumption; hence why they are only visible when app.debug = true. Displaying them anyways in production is a bad idea for security as it reveals vital information to attackers about the state of the user database and how close they are to getting in
- The Storm library cannot reference language keys provided by the Backend module, that is coupling the separate codebases in the wrong direction and would cause issues
- If we were to even consider implementing this it would need to be by first implementing more specific AuthenticationException classes and throwing them when applicable and then in the backend module catching each specific exception class to provide the translated error message for that specific exception class.

I'm going to close this and related PRs for now, but feel free to post your thoughts in response to this message and we can continue the discussion and reopen the PRs if necessary.

PR #118, opened 2022-09-27
Slug generation performance improvement

RomainMazB:
The `EXISTS` sql statement will return true at the first existing value where the `COUNT` will uselessly continue to loop through the whole table which is inaccurate for this use case.

The syntax is more readable too IMO.

### DIFF
--- a/src/Database/Traits/Sluggable.php
+++ b/src/Database/Traits/Sluggable.php
@@ -95,8 +95,8 @@ trait Sluggable
         $separator = $this->getSluggableSeparator();
         $_value = $value;
         while (($this->methodExists('withTrashed') && $this->allowTrashedSlugs) ?
-            $this->newSluggableQuery()->where($name, $_value)->withTrashed()->count() > 0 :
-            $this->newSluggableQuery()->where($name, $_value)->count() > 0
+            $this->newSluggableQuery()->where($name, $_value)->withTrashed()->exists() :
+            $this->newSluggableQuery()->where($name, $_value)->exists()
         ) {
             $counter++;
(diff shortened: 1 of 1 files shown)


bennothommo:
I've given this a test by running the following in the Sluggable tests:

```php
public function testSlugGenerationLargeDataset()
{
    // Create 2500 records
    for ($i = 0; $i < 2500; ++$i) {
        TestModelSluggable::create([
            'name' => 'test row ' . $i,
        ]);
    }

    $testModel1 = TestModelSluggable::create(['name' => 'test']);

    // Create another 2500 records
    for ($i = 2500; $i < 5000; ++$i) {
        TestModelSluggable::create([
            'name' => 'test row ' . $i,
        ]);
    }

    DB::listen(function ($query) {
        print_r([
            'query' => $query->sql,
            'bindings' => $query->bindings,
            'time' => $query->time,
        ]);
    });

    $testModel2 = TestModelSluggable::create(['name' => 'test']);
}
```

I found a very, very, very small performance benefit for using `exists()` over `count()`. I mean incredibly minuscule - even with 5000 records. `exists()` ran in 0.05 seconds, `count()` ran in 0.06 seconds, but it fluctuated, so I assume there's rounding going on. This is with SQLite.

The logic is sound, so I'm not against merging it - it makes sense that we only care if a record exists with a given slug, we don't need to know *how many* records have the slug, but I just wanted to make it clear this is a micro-optimisation.

LukeTowers:
While the gains are small, the change does take it from an issue that gets slower with scale and turns it into a consistent performance operation instead which is always better IMO.

PR #128, opened 2022-10-21
Fix model.relation.afterAttach event handler

mjauvin:
Take into account $id param of `model.relation.afterAttach` could be a model


bennothommo:
@mjauvin is this still in draft?

mjauvin:
Nope!

PR #130, opened 2022-10-24
Prioritize local dynamic methods over behavior-provided methods

LukeTowers:
This prioritizes locally added dynamic methods over public methods that have been defined in behaviors that are implemented on the object.

### DIFF
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -424,27 +424,31 @@ trait ExtendableTrait
     }
 
     /**
-     * Magic method for `__call()`
+     * Magic method for `__call()`. Priority is as follows:
+     * - "Dynamic Methods" added locally to the object via addDynamicMethod($name, $callable)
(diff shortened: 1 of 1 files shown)


src/Extension/ExtendableTrait.php
```diff
@@ -424,13 +424,25 @@ public function extendableSet($name, $value)
     }
 
     /**
-     * Magic method for `__call()`
+     * Magic method for `__call()`. Priority is as follows:
```
bennothommo:
```suggestion
     * Magic method for `__call()`.
     *
     * Callback priority is as follows:
```

bennothommo (approved):

mjauvin:
I've been dreaming of this for the past 6 years!

LukeTowers:
It's still not going to help with the problem of overriding methods that already exist on the class or its ancestors directly but at least now we can override methods that are defined on the behaviors

PR #133, opened 2022-11-22
Convert Markdown parser to CommonMark

bennothommo:
Currently, our Markdown Parser uses the [`erusev/parsedown`](https://github.com/erusev/parsedown) and [`erusev/parsedown-extra`](https://github.com/erusev/parsedown-extra) libraries for Markdown parsing.

Both libraries have not seen updates for a number of years (beyond README updates and a proposed 2.0 branch), and while they state they follow the GitHub-Flavoured Markdown specification, there are subtle incompatibilities with the spec. This PR proposes a move to [`league/commonmark`](https://github.com/thephpleague/commonmark), which is already a dependency from Laravel, thus dropping a dependency from the Storm library.

Beyond the fact that the CommonMark library is more readily maintained, it also adheres to the [CommonMark specification](https://spec.commonmark.org/), making its functionality documented and expected, and also adheres to the GitHub-Flavoured Markdown specification.

It is also incredibly customisable and extendable, and allows us to tailor an API for the Markdown parser to allow people to customise what they want to allow in Markdown parsing, whilst still maintaining parity with what people expect from the old API (which is still very much supported). I have imported the test cases from the Parsedown library to demonstrate that the proposed switchover should not have an effect on most usage.

This will probably require some slight adjustments to the Blog and Forum plugins (and the Docs plugin, which spurred this PR), but for the most part, I believe this should be a drop-in replacement.


src/Parse/Markdown/CommonMarkCoreExtension.php
```diff
@@ -0,0 +1,174 @@
...
+use League\CommonMark\Environment\EnvironmentBuilderInterface;
+use League\CommonMark\Extension\CommonMark\Delimiter\Processor\EmphasisDelimiterProcessor;
+use League\CommonMark\Extension\ConfigurableExtensionInterface;
+use League\CommonMark\Node as CoreNode;
+use League\CommonMark\Parser as CoreParser;
+use League\CommonMark\Renderer as CoreRenderer;
+use League\Config\ConfigurationBuilderInterface;
+use Nette\Schema\Expect;
+
+final class CommonMarkCoreExtension implements ConfigurableExtensionInterface
```
LukeTowers: Does this have to be final? What's the benefit?

bennothommo: Honestly, it was just a copy-paste from the original extension from the Commonmark library. I'll remove the `final` anyway.

LukeTowers (approved):
LGTM!

src/Support/Facades/Markdown.php
```diff
@@ -3,19 +3,56 @@
...
  */
 class Markdown extends Facade
 {
     /**
-     * Get the registered name of the component.
-     *
-     * @return string
+     * {@inheritDoc}
+     */
+    protected static $cached = true;
```
bennothommo:
```suggestion
    protected static $cached = false;
```

PR #134, opened 2022-12-08
Allow scoped and local extensions

bennothommo:
This allows one to create either a scoped global extension through `static::extend()`, which will apply to all newly-constructed instances of a class going forward, or they can also use `$object->extend()` to apply an extension to the single instance at run-time.

This allows access to protected and private methods within an instance, allowing full customisation of a given instance if it's allowed to be extendable.

### DIFF
--- a/.github/workflows/code-analysis.yaml
+++ b/.github/workflows/code-analysis.yaml
@@ -44,6 +44,7 @@ jobs:
         with:
           php-version: '8.0'
           extensions: ${{ env.extensions }}
+          coverage: none
 
       - name: Setup dependency cache
         id: composercache
--- a/.github/workflows/code-analysis.yaml
+++ b/.github/workflows/code-analysis.yaml
@@ -28,7 +28,7 @@ jobs:
         id: extcache
         uses: shivammathur/cache-extensions@v1
         with:
-          php-version: '8.0'
+          php-version: '8.1'
           extensions: ${{ env.extensions }}
           key: ${{ env.key }}
 
@@ -42,7 +42,7 @@ jobs:
       - name: Install PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.0'
+          php-version: '8.1'
           extensions: ${{ env.extensions }}
           coverage: none
 
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -22,6 +22,7 @@ jobs:
         with:
           php-version: '7.3'
           tools: phpcs
+          coverage: none
       - name: Run code quality checks (on push)
         if: github.event_name == 'push'
         run: ./.github/workflows/utilities/phpcs-push ${{ github.sha }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,6 +50,7 @@ jobs:
         with:
           php-version: ${{ matrix.phpVersion }}
           extensions: ${{ env.extensions }}
+          coverage: none
 
       - name: Setup dependency cache
         id: composercache
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Winter Storm
 =======
 
-[![Tests](https://img.shields.io/github/workflow/status/wintercms/storm/Tests/develop?label=tests&style=flat-square)](https://github.com/wintercms/storm/actions)
+[![Tests](https://img.shields.io/github/actions/workflow/status/wintercms/storm/tests.yml?branch=develop&label=tests&style=flat-square)](https://github.com/wintercms/storm/actions)
 [![License](https://img.shields.io/github/license/wintercms/storm?label=open%20source&style=flat-square)](https://packagist.org/packages/winter/storm)
 [![Discord](https://img.shields.io/discord/816852513684193281?label=discord&style=flat-square)](https://discord.gg/D5MFSPH6Ux)
 
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
 
         "assetic/framework": "~3.0",
         "doctrine/dbal": "^2.6",
-        "enshrined/svg-sanitize": "^0.15",
+        "enshrined/svg-sanitize": "~0.16",
         "laravel/framework": "^9.1",
         "laravel/tinker": "^2.7",
         "league/csv": "~9.1",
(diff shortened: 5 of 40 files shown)


LukeTowers:
Could you add some usage examples?

bennothommo:
@LukeTowers sure, so one example, which spurned this PR, was me trying to emulate the `run` method in the `Backend\Classes\Controller` method - specifically, this block: (https://github.com/wintercms/winter/blob/develop/modules/backend/classes/Controller.php#L219-L243) - within a behaviour.

Since the behavior is provided the controller as a `$controller` property, I can emulate anything that uses public methods and properties, but I can't use the protected `$publicActions` and `$requiredPermissions` properties of that controller.

For example - this won't work:

```php
    public function apiRun(string $action, array $params = [])
    {
        $this->controller->action = $action;
        $this->controller->params = $params;
        $this->controller->suppressView = true;

        // Determine if the action is public
        $isPublicAction = in_array($action, $this->controller->publicActions);

        if (!$isPublicAction) {
            if ($this->controller->requiredPermissions && !$this->controller->user->hasAnyAccess($this->controller->requiredPermissions)) {
                return $this->errorResponse('Unable to access this API endpoint', 403);
            }
        }
```

With this PR in place, I can use a local extension and return the callback to determine authorisation:

```php
    public function apiRun(string $action, array $params = [])
    {
        $this->controller->action = $action;
        $this->controller->params = $params;
        $this->controller->suppressView = true;

        // Determine if the action is public
        $allowed = $this->controller->extend(function () {
            $isPublicAction = in_array($action, $this->publicActions);

            if (!$isPublicAction) {
                if ($this->requiredPermissions && !$this->user->hasAnyAccess($this->requiredPermissions)) {
                    return false;
                }
            }
           
            return true;
        });

        if (!$allowed) {
              return $this->errorResponse('Unable to access this API endpoint', 403);
        } 
```

bennothommo:
Another example, say I wanted to make a global extension, but needed access to some protected methods or properties, this would now be possible:

```php
FormWidgetBase::extend(function () {
    // $valueFrom is protected, and has no getter in this class.
    if ($this->valueFrom === 'my_field') {
        $this->addDynamicMethod('someMethod', function () {
            // ....
        });
    }
}, true);
```

bennothommo:
Just realised in the most recent tests that this also means that we can now define dynamic methods that can use protected or private methods and properties within.

LukeTowers:
Hmm, with that example I'm not so sure that extend is the right name for that method, at least in the context of the use case of injecting the provided closure into the context of the object instance that `extend()` is being called on. Is that the only use case for that method or would there be other reasons to use it like that? Also what would `->extend()` inside of an `::extend()` look like / be used for / actually do?

bennothommo:
@LukeTowers the name can certainly be worked on if there's any better ideas, although `extend` does hew close to class `extends`, which is what this functionality does on a smaller scale IMO.

My first example is an example of a global extension (ie. `Class:extend()`) moving through to a local extension (ie. `$class->extend`), since it's a snippet of a controller behavior that is adding an `apiRun` method that then needs to extend further into the controller that the behavior is attached to. Effectively, the local extension is just acting on the specific controller as opposed to acting on *any* controller.

bennothommo:
@LukeTowers any further thoughts on the above?

mjauvin:
Tested this with my use case where I need to have access to methods added by behavior and it works beautifully:
```
MyModel::extend(function () {
    $this->extend(function () {
        $this->behaviorMethodAvailable();
    });
}, true);
```

And it's amazingly nice to be able to use `$this` for both static and local closures!

mjauvin:
This is so freaking powerful!!! This changes everything.

Amazing work @bennothommo!

mjauvin:
What's also very nice is behavior is unchanged if `Class::extend()` is called only with a $callback.
Also, the error is much more clear when no $callback is provided:

before:
```
\Cms\Classes\Page::extend();

   ArgumentCountError  Too few arguments to function Winter\Storm\Extension\Extendable::extend(), 0 passed in /home/w2c/w2c.studioazura.comeval()'d code on line 1 and exactly 1 expected.
```
after:
```
\Cms\Classes\Page::extend();

   InvalidArgumentException  The extend() method requires a callback parameter or closure.
```

LukeTowers:
Let's get the conflicts resolved and a quick docs PR up and then I can review

bennothommo:
Ready to go @LukeTowers

mjauvin (approved):
Looks good, can't wait to use this!

mjauvin:
> Tested this with my use case where I need to have access to methods added by behavior and it works beautifully:
> 
> ```
> MyModel::extend(function () {
>     $this->extend(function () {
>         $this->behaviorMethodAvailable();
>     });
> }, true);
> ```

@bennothommo I successfully tested this before the PR got merged, but now it doesn't see the behavior methods...

bennothommo:
@mjauvin I'd need to see the context of how you've implemented the extension.

mjauvin:
@bennothommo here is an example:
```php
public function boot(): void
{
    \StudioAzura\Aerocom\Models\Service::extend(function () {
        $this->extend(function () {
            $this->addTranslatableAttributes('myTranslatableField');
        });
    }, true);
}
```

The Service model implements the TranslatableModel behavior and should have the `addTranslatableAttributes()` method available at this point, but I get error:


Call to undefined method StudioAzura\Aerocom\Models\Service::addTranslatableAttributes()

I tested this exact case before the PR got merged, so I know it was working.

What am I missing ?

bennothommo:
@mjauvin I'm not sure how it worked before 😕. Basically, the extension callbacks (ie. the static `extend` call) are called *before* the behaviors are loaded. All this PR did was enable the ability to use protected and private methods - it didn't change the ordering of initialisation.

bennothommo:
Wait - I saw my comment in your PR that was to introduce the `after` option to run callbacks after the behaviors are loaded - your way above should work. I'll investigate this further.

bennothommo:
@mjauvin This commit should fix it (a8ba4eef2e96851c5f2293f1fbf70047c3004159), however, I'm still curious as to how it worked before, because it should never have. Static `extend` calls are always fired before behaviors are loaded, because they provide the ability to dynamically add behaviors. Local extensions (ie `$object->extend`) fire immediately on the given object, which means when used in a static `extend` call, it would've fired before the behaviors too.

The commit above will delay local extensions until after the construction is complete, because that was always the intention of them to be available on "constructed" objects (like when used in run-time).

mjauvin:
I don't know, but I'm 99% sure I tested this when I wrote the comment in this PR. Maybe I did something stupid thinking it worked

mjauvin:
@bennothommo the problem is still there, behavior method is still not found.

bennothommo:
@mjauvin are you sure? The new test case I introduced in that commit is testing your exact scenario?

mjauvin:
@bennothommo pretty sure. Where is `getFoo()` defined ? Is it in a behavior implemented by the model ?

mjauvin:
@bennothommo problem is your test class extends the Extendable class, mine extends the Model class (which uses the ExtendableTrait)

PR #135, opened 2022-12-09
Ensure that getAvailablePaths respects the maxDepth of the file iterator

jaxwilko:
This PR fixes an issue brought to light by https://github.com/wintercms/winter/pull/791. 

When `FileDatasource::getAvailablePaths()` is called, it does not respect the same maxDepth setting that is set in `FileDatasource::select()` meaning that the reported paths differ from the selectable paths.

This PR adds the `setMaxDepth()` call on the iterator used in `getAvailablePaths()` and sets a property for `maxDepth` to remove magic numbers and allow for custom extensions.

### DIFF
--- a/src/Halcyon/Datasource/FileDatasource.php
+++ b/src/Halcyon/Datasource/FileDatasource.php
@@ -36,8 +36,6 @@ class FileDatasource extends Datasource
 
     /**
      * The maximum depth of the filesystem to scan, defaulting to 1 subdirectory.
-     *
-     * @var int
      */
     protected int $maxDepth = 1;
 
--- a/src/Halcyon/Datasource/FileDatasource.php
+++ b/src/Halcyon/Datasource/FileDatasource.php
@@ -35,6 +35,13 @@ class FileDatasource extends Datasource
     protected $resolvedBasePaths = [];
 
     /**
+     * The maximum depth of the filesystem to scan, defaulting to 1 subdirectory.
+     *
+     * @var int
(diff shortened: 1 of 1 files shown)


src/Halcyon/Datasource/FileDatasource.php
```diff
@@ -34,6 +34,13 @@ class FileDatasource extends Datasource
      */
     protected $resolvedBasePaths = [];
 
+    /**
+     * The maximum depth of the filesystem to scan, defaulting to 1 subdirectory.
+     *
+     * @var int
```
LukeTowers:
```suggestion
```

src/Halcyon/Datasource/FileDatasource.php
```diff
@@ -337,6 +342,10 @@ public function getAvailablePaths(): array
             ? new RecursiveIteratorIterator(new RecursiveDirectoryIterator($this->basePath))
             : [];
 
+        if (!is_array($it)) {
+            $it->setMaxDepth($this->maxDepth + 1);
+        }
```
LukeTowers, line 347: What does this do?

jaxwilko: Sets the iterator maxDepth on the iterator used inside `getAvailablePaths()`. This is to ensure that the reported paths match the same reported paths in `select()` as both methods implement their own iterators

LukeTowers (approved):

PR #137, opened 2023-01-10
Laravel 10 support

bennothommo:
WIP PR for Laravel 10 support.


composer.json
```diff
@@ -36,10 +36,10 @@
         "ext-zip": "*",
 
         "assetic/framework": "~3.0",
-        "doctrine/dbal": "^2.6",
-        "enshrined/svg-sanitize": "~0.16",
-        "laravel/framework": "^9.1",
-        "laravel/tinker": "^2.7",
+        "doctrine/dbal": "^3.0",
+        "enshrined/svg-sanitize": "^0.15",
```
LukeTowers: Why was this bumped down?

bennothommo: Must've missed this in the branch merge. Will adjust.

bennothommo:
@LukeTowers regarding the `tags` method in the Cache, according to [Laravel's docs](https://laravel.com/docs/10.x/cache#cache-tags), it's not supported on some cache types, so type hinting to the contract is correct. I've scouted both Winter and Storm and I don't see it actually being used, so I wonder if we just drop it from Halcyon?

LukeTowers:
@bennothommo all of the core cache repositories provided by Laravel extend the base Repository class, even the ones that don't support tags, so I'm fine with type hinting it instead.

PR #141, opened 2023-02-21
Initial WIP on improving the File model to use more native disk methods

LukeTowers:
WIP


src/Database/Attach/File.php
```diff
@@ -271,10 +257,8 @@ public function setDataAttribute($value)
      * Helper attribute for get image width.
      *
      * Returns `null` if this file is not an image.
-     *
-     * @return string|int|null
      */
-    public function getWidthAttribute()
+    public function getWidthAttribute(): string|int|null
```
Denoder, line 268: Could you consider changing the width and height attribute to be optional when using an s3 driver in some form. These 2 attributes make any remote storage take more than 7 seconds because it has to download the file locally.

Denoder: actually nvm, I can just override it

LukeTowers: We could store that data on upload theoretically and retrieve it from metadata instead of recalculating

Denoder: hmm, if that's the case, wouldn't it also be beneficial to have a `has_thumb` meta parameter to indicate if the image has a thumbnail or not.

Currently the way I have it is that I have a `has_thumb` property on the file model. so rather than using `hasFile` which checks the remote file location it checks if that parameter is set, then I offload the thumbnail generation to a queue (this is more of a custom solution for me that im utilizing, but using `hasFile` slows down the process a lot.)

LukeTowers: We should probably add a `metadata` column in the form of a JSON blob to store this sort of information

LukeTowers: @Denoder added the metadata column in https://github.com/wintercms/storm/commit/d44beb398ecaee8db39e6f2a7a2a7404f5d6996c

mjauvin:
@LukeTowers has this been tested extensively?

LukeTowers:
@mjauvin should be minimal impact, just requires the other PR that I'm finalizing right now for the metadata support

mjauvin:
> @mjauvin should be minimal impact, just requires the other PR that I'm finalizing right now for the metadata support

I would feel a lot more confident with some unit tests, this is used extensively in my projects

LukeTowers:
@mjauvin see https://github.com/wintercms/winter/commit/94ad5e85b0216c0f93516e3bd6fb04cea55c5e2f for the changes required on the system module. These changes are relatively minor and should be non-breaking; but the changes in https://github.com/wintercms/winter/pull/844 are going to need some more testing

mjauvin:
@LukeTowers I get the following when updating to latest dev-develop:
```
In File.php line 48:
                                                                                                                                                              
  Declaration of System\Models\File::getPath($fileName = null) must be compatible with Winter\Storm\Database\Attach\File::getPath(?string $fileName = null):  
   string in /var/www/vhosts/wn.r4l.com/public_html/modules/system/models/File.php on line 48                                                                                                                                                     
```

Oh, never mind, I had not updated the system module...

LukeTowers:
All good now @mjauvin?

PR #143, opened 2023-03-01
Prevent empty id attribute

mjauvin:
Replaces #142

Add unit tests

### DIFF
--- a/src/Html/FormBuilder.php
+++ b/src/Html/FormBuilder.php
@@ -285,6 +285,10 @@ class FormBuilder
             // when creating the HTML element. Then, we will return the entire input.
             $merge = compact('type', 'value', 'id');
 
+            if ($id === "") {
+                unset($merge['id']);
+            }
+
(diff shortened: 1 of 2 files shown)


mjauvin:
You were right @LukeTowers, using isset() was not correct in this case.

This version resolves the issue.

AIC-BV:
Thanks @mjauvin

mjauvin:
Does that PR solve your issue @AIC-BV ?

AIC-BV:
> Does that PR solve your issue @AIC-BV ?

Yes, it removes all these warnings, thus removing the w3c errors
image

mjauvin:
It's not clear from the screenshot you referenced above... glad to hear it works!

AIC-BV:
> It's not clear from the screenshot you referenced above... glad to hear it works!

Well the screenshot is the before, there is no after because it's empty😂

LukeTowers:
@bennothommo can we fix the code analysis tests?

LukeTowers (approved):

PR #144, opened 2023-04-03
Fix missing attributes in FormBuilder inputs

### LINKED ISSUE #864: FormBuilder form_submit() not working as expected
### Winter CMS Build

dev-develop

### PHP Version

8.1

### Database engine

SQLite

### Plugins installed

_No response_

### Issue description

First of all, maybe this is not issue/bug but is related to https://github.com/wintercms/meta/blob/master/l9-upgrade-notes.md#html changes..., although clarification there suggests that it applies more for extended plugin classes than `form_...` calls in Twig templates.

Anyways, in WinterCMS 1.1.8 it was possible to use `{{ form_submit('Apply', {'class': 'btn btn-primary'}) }}` and render valid submit button with following output `<input class="btn btn-primary" type="submit" value="Apply">` but after upgrading to WinterCMS 1.2.x-dev the output is only `<input class="btn btn-primary">`. In other words, submit type of input field and also value passed with the first parameter is missing in the rendered output.

Looks like issue was introduced with https://github.com/wintercms/storm/commit/2ef32346682c768a9d457c4dba6796f34b644eab#diff-ec32bc4f8918dddd6340e4e341849935c6765bdfba774c75b7553376a6109e2eR274 (@bennothommo)

### Steps to replicate

In Twig file (page or component .html) create form and use `{{ form_submit('Apply', {'class': 'btn btn-primary'}) }}` or reduced version of that - `{{ form_submit('Apply') }}`. The rendered input field will not have submit type and also passed value "Apply" is missing (from https://wintercms.com/docs/services/html#buttons it's expected that the passed value is text which needs to be shown on the submit button).

### Workaround

As an alternative solution it is possible to use:

- `{{ form_submit('Apply', {'class': 'btn btn-primary', 'type': 'submit', 'value': 'Apply'}) }}` or
- `{{ form_button('Apply', {'class': 'btn btn-primary', 'type': 'submit'}) }}`

bennothommo:
Commit https://github.com/wintercms/storm/commit/2ef32346682c768a9d457c4dba6796f34b644eab made some adjustments to the input method arguments that resulted in the some attributes not being populated if the name of the input was not provided. This commit changes the structure of the condition so that these attributes are properly filled in.

I've also redone the tests using some custom assertions so that the order of the attributes does not matter when testing. The previous tests were a bit flimsy in this regard.

Fixes https://github.com/wintercms/winter/issues/864

### DIFF
--- a/src/Html/FormBuilder.php
+++ b/src/Html/FormBuilder.php
@@ -30,7 +30,7 @@ class FormBuilder
     /**
      * The session store implementation.
      */
-    protected ?\Illuminate\Session\Store $session;
+    protected ?\Illuminate\Session\Store $session = null;
 
     /**
      * The current model instance for the form.
@@ -228,9 +228,9 @@ class FormBuilder
     {
         $token = !empty($this->csrfToken)
             ? $this->csrfToken
-            : $this->session->token();
+            : ($this->session?->token() ?? null);
 
-        return $this->hidden('_token', $token);
+        return ($token) ? $this->hidden('_token', $token) : '';
     }
 
     /**
@@ -270,29 +270,28 @@ class FormBuilder
             $options['name'] = $name;
         }
 
+        $merge = [
+            'type' => $type,
+        ];
+
         if (!empty($name)) {
             // We will get the appropriate value for the given field. We will look for the
             // value in the session for the value in the old input data then we'll look
             // in the model instance if one is set. Otherwise we will just use empty.
-            $id = $this->getIdAttribute($name, $options);
+            $merge['id'] = $this->getIdAttribute($name, $options);
 
-            if (!in_array($type, $this->skipValueTypes)) {
-                $value = $this->getValueAttribute($name, $value);
-            }
-
-            // Once we have the type, value, and ID we can merge them into the rest of the
-            // attributes array so we can convert them into their HTML attribute format
-            // when creating the HTML element. Then, we will return the entire input.
-            $merge = compact('type', 'value', 'id');
-
-            if ($id === "") {
+            if ($merge['id'] === "") {
                 unset($merge['id']);
             }
-
-            $options = array_filter(array_merge($options, $merge), function ($item) {
-                return !is_null($item);
-            });
         }
+
+        if (!in_array($type, $this->skipValueTypes)) {
+            $merge['value'] = $this->getValueAttribute($name, $value);
+        }
+
+        $options = array_filter(array_merge($options, $merge), function ($item) {
+            return (!is_null($item) && $item !== false);
+        });
 
         return '<input' . $this->html->attributes($options) . '>';
     }
@@ -849,11 +848,11 @@ class FormBuilder
     /**
      * Get the value that should be assigned to the field.
      *
-     * @param  string  $name
-     * @param  string|array  $value
+     * @param  string|null  $name
+     * @param  string|array|int|null  $value
      * @return string|array|null
      */
-    public function getValueAttribute($name, $value = null)
+    public function getValueAttribute(?string $name = null, $value = null)
     {
         if (empty($name)) {
             return $value;
--- a/src/Html/FormBuilder.php
+++ b/src/Html/FormBuilder.php
@@ -279,10 +279,6 @@ class FormBuilder
             // value in the session for the value in the old input data then we'll look
             // in the model instance if one is set. Otherwise we will just use empty.
             $merge['id'] = $this->getIdAttribute($name, $options);
-
-            if ($merge['id'] === "") {
-                unset($merge['id']);
-            }
         }
 
         if (!in_array($type, $this->skipValueTypes)) {
@@ -830,7 +826,7 @@ class FormBuilder
      *
      * @param  string  $name
      * @param  array   $attributes
-     * @return string
+     * @return string|null
      */
     public function getIdAttribute($name, $attributes)
     {
@@ -842,7 +838,7 @@ class FormBuilder
             return $name;
         }
 
-        return '';
+        return null;
     }
 
     /**
--- a/tests/Html/FormBuilderTest.php
+++ b/tests/Html/FormBuilderTest.php
@@ -232,6 +232,16 @@ class FormBuilderTest extends TestCase
         $this->assertElementAttributeEquals('value', 'my value', $result);
     }
 
+    public function testFormInputBooleanAttribute()
+    {
+        $result = $this->formBuilder->input(type: 'text', name: 'my-name', value: 'my value', options: ['required']);
+
+        $this->assertElementIs('input', $result);
+        $this->assertElementAttributeEquals('name', 'my-name', $result);
+        $this->assertElementAttributeEquals('type', 'text', $result);
+        $this->assertElementHasAttribute('required', $result);
+    }
+
     /**
      * @testdox can create a text input of type "email".
      */
--- a/tests/Html/FormBuilderTest.php
+++ b/tests/Html/FormBuilderTest.php
@@ -6,11 +6,25 @@ use Winter\Storm\Router\UrlGenerator;
 
 use Illuminate\Http\Request;
 use Illuminate\Routing\RouteCollection;
+use Winter\Storm\Tests\Assertions\AssertHtml;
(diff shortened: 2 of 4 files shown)


src/Html/FormBuilder.php
```diff
@@ -849,11 +848,11 @@ public function getIdAttribute($name, $attributes)
...
      * Get the value that should be assigned to the field.
      *
-     * @param  string  $name
-     * @param  string|array  $value
+     * @param  string|null  $name
+     * @param  string|array|int|null  $value
      * @return string|array|null
      */
-    public function getValueAttribute($name, $value = null)
+    public function getValueAttribute(?string $name = null, $value = null)
```
mjauvin, line 851: @bennothommo The only way I see the `$name` argument not being provided is by calling the method without arguments or calling it with `$value` as a named argument by itself. Is that why you added a null default value ?

src/Html/FormBuilder.php
```diff
@@ -270,30 +270,29 @@ public function input(string $type, ?string $name = null, ?string $value = null,
...
-            }
+            $merge['id'] = $this->getIdAttribute($name, $options);
 
-            // Once we have the type, value, and ID we can merge them into the rest of the
-            // attributes array so we can convert them into their HTML attribute format
-            // when creating the HTML element. Then, we will return the entire input.
-            $merge = compact('type', 'value', 'id');
-
-            if ($id === "") {
+            if ($merge['id'] === "") {
```
mjauvin: Seems a bit pointless to assign and unassign now that the `compact()` method is not being called.

tests/Html/FormBuilderTest.php
```diff
@@ -19,40 +33,280 @@ public function setUp() : void
...
+
+    /**
+     * @testdox ignores an ID that is boolean "false".
+     */
+    public function testFormInputTextFalse()
+    {
+        $result = $this->formBuilder->input(type: 'text', name: 'my-name', value: 'my value', options: ['id' => false]);
+
+        $this->assertElementIs('input', $result);
+        $this->assertElementDoesntHaveAttribute('id', $result);
```
mjauvin, line 215: Previous behavior was different, id should be present with value as empty string.

mjauvin (commented):
Great work on the AssertHtml trait @bennothommo!

Just a few things I noticed in the new test cases.

src/Html/FormBuilder.php
bennothommo: @mjauvin no, it was specifically to ensure that the value is populated in the `input` method in all scenarios - for example, the `submit` method usage that spurned this whole PR in the first place. @arvislacis had a `submit` call which had a value but no name for the field, so when it is passed to this method, `name` is null but the value is not, so it returns the value.

src/Html/FormBuilder.php
bennothommo: It's to ensure that the ID has a value - other attributes may allow empty strings (eg. `required` or data attributes) but if the ID is an empty string, it's removed.

tests/Html/FormBuilderTest.php
bennothommo: This is testing `false` though? `false` means no attribute.

bennothommo:
Thanks for reviewing @mjauvin :)

src/Html/FormBuilder.php
mjauvin: Yes, I understand, I'm the one that made that fix previously... what I mean is this test should now be done in `getIdAttribute()` method.

src/Html/FormBuilder.php
mjauvin: Not sure I understand your point, the following call should behave the same whether the method signature has "$name = null" or not, shouldn't it ?
```
getValueAttribute(null, "a Value");
```

tests/Html/FormBuilderTest.php
mjauvin: I'm just saying this changes the previous behavior (which my initial tests were replicating). The idea was to NOT change the original resulting behavior, otherwise hell will break loose... ;)

src/Html/FormBuilder.php
bennothommo: Yes, but PHPStan would whinge about it, because of [this block of code](https://github.com/wintercms/storm/pull/144/files#diff-ec32bc4f8918dddd6340e4e341849935c6765bdfba774c75b7553376a6109e2eL279-L281) being moved outside of the condition checking if `$name` is not empty.

mjauvin: Ahh, gotcha!

src/Html/FormBuilder.php
bennothommo: Fair point, I will change that.

tests/Html/FormBuilderTest.php
bennothommo: Hmmm, might need to add some more unit tests for this then, because I feel that if someone were to add *any* attribute with `false` (or `null`) as the value, that's effectively saying that they *don't* want that attribute to be visible.

For example, someone adds the following:

```
$this->formBuilder->input('text', 'field-name', 'field-value', [
    'required' => false,
]);
```

The field should *not* contain a `required` attribute - if it were to be coerced to an empty string, that would still make the field required by HTML conventions.

bennothommo: (I know this sounds like me not following my own rules about BC breaks, but they're acceptable if they fix a larger issue IMO :stuck_out_tongue: )

mjauvin: I feel the same, but tested this before making the test cases and that was the behevior before the changes I made. So unless your changes with the `if (!empty($name))` broke the behavior at the time, this is going to be a breaking change.

arvislacis: Check also https://github.com/wintercms/storm/blob/674733f16629e86cf7f577f29f26fa8c2a0d69c5/src/Html/FormBuilder.php#L693 and https://github.com/wintercms/storm/blob/674733f16629e86cf7f577f29f26fa8c2a0d69c5/src/Html/FormBuilder.php#L695 `return $this->input('submit', null, $value, $options);` - for submit button null value is passed as a name.

bennothommo: So I've been able to confirm that before my fixes, using the following code:

```php
$this->formBuilder->input('text', 'field-name', 'field-value', [
    'required' => false,
]);
```

generates:

```html
<input type="text" name="field-name" value="field-value" required="">
```

Which, even though the dev has explicitly said that required is `false`, still would make the field required in the browser. With my fixes in place, the following is generated:

```html
<input type="text" name="field-name" value="field-value">
```

Which I feel is correct. When the required value is `true`, both the original code and the revised code show the following:

```html
<input type="text" name="field-name" value="field-value" required="1">
```

The original code had made an allowance that if one were to use an integer key, then it adds a boolean attribute like this:

```php
$this->formBuilder->input('text', 'field-name', 'field-value', [
    'required',
]);
```

```html
<input type="text" name="field-name" value="field-value" required>
```

In balance, I think a BC break might be necessary here.

@LukeTowers @jaxwilko any thoughts on this?

LukeTowers: @bennothommo I'm fine with the BC break. If you're trying to set false or null on an HTML attribute value then you can use the string literal representation of it, so long as we allow for an empty string to be a valid value I'm happy with it.

src/Html/FormBuilder.php
LukeTowers: @bennothommo is this still a change you're going to make?

tests/Html/FormBuilderTest.php
bennothommo: Alright, we're going to treat this as a BC break. Boolean attributes still work as previously defined (ie. adding `['required']` to `$options`).

src/Html/FormBuilder.php
bennothommo: Done. :)

mjauvin (approved):
Looks good to me, with all those amazing tests!

LukeTowers (approved):

PR #147, opened 2023-05-09
Add json helper method to set payload as json-encoded value.

mjauvin:
Allows doing this:
```php
$payload = [
    'key1' => 'value1',
    'key2' => 'value2'
];

$apiResult = Http::post($url, function ($http) use ($payload) {
    $http->json( $payload );
});

```
Instead of this:
```php
$payload = [
    'key1' => 'value1',
    'key2' => 'value2'
];

$apiResult = Http::post($url, function ($http) use ($payload) {
    $http->header('Content-Type', 'application/json; charset=utf-8');
    $http->setOption( CURLOPT_POSTFIELDS, json_encode($payload) );
```

### DIFF
--- a/src/Network/Http.php
+++ b/src/Network/Http.php
@@ -449,7 +449,7 @@ class Http
     {
         if (!Str::isJson($payload)) {
             if (!$payload = json_encode($payload)) {
-                throw new ApplicationException('provided payload cannot be json_encoded ');
+                throw new ApplicationException('The provided payload failed to be encoded as JSON');
             }
         }
 
--- a/src/Network/Http.php
+++ b/src/Network/Http.php
@@ -115,7 +115,7 @@ class Http
     public $requestOptions = [];
 
     /**
-     * @var array Request data.
+     * @var array|string Request data.
      */
     public $requestData;
 
@@ -440,6 +440,20 @@ class Http
             }
         }
         return $headers;
+    }
+
+    /**
+     * Add JSON encoded payload
+     */
+    public function json(array|string $payload) : self
+    {
+        if (is_array($payload)) {
+            $payload = json_encode($payload);
+        }
+        $this->requestData = $payload;
+        $this->header('Content-Type', 'application/json');
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Network/Http.php
```diff
@@ -441,6 +442,20 @@ protected function headerToArray($header)
         return $headers;
     }
 
+    /**
+     * Add JSON encoded payload
+     */
+    public function json(array|string $payload) : self
```
bennothommo:
```suggestion
    public function json(array|string $payload): self
```

src/Network/Http.php
```diff
@@ -452,10 +467,11 @@ public function data($key, $value = null)
             foreach ($key as $_key => $_value) {
                 $this->data($_key, $_value);
             }
-            return $this;
+        } elseif (is_null($value) && Str::isJson($key)) {
+            $this->requestData = $key;
+        } else {
+            $this->requestData[$key] = $value;
```
bennothommo: Now that we have the `json()` helper method, I don't see a need for this method to change.

mjauvin: removed

mjauvin:
Any objection to merge this one ?

src/Network/Http.php
```diff
@@ -441,21 +442,32 @@ protected function headerToArray($header)
         return $headers;
     }
 
+    /**
+     * Add JSON encoded payload
+     */
+    public function json(array|string $payload): self
```
LukeTowers: json_encode can take a lot more types than just array|string, we should probably support everything that they do. To check to see if it needs to be encoded or not we can just pass it to Str::isJson()

src/Network/Http.php
```diff
@@ -441,21 +442,35 @@ protected function headerToArray($header)
...
     }
 
+    /**
+     * Add JSON encoded payload
+     */
+    public function json(mixed $payload): self
+    {
+        if (!Str::isJson($payload)) {
+            if (!$payload = json_encode($payload)) {
+                throw new ApplicationException('provided payload cannot be json_encoded ');
```
LukeTowers:
```suggestion
                throw new ApplicationException('The provided payload failed to be encoded as JSON');
```

PR #149, opened 2023-07-21
Add replicating and pruning event handlers

marvindurot:
This PR adds nice event handlers for `replicating` and `pruning` Laravel events in the Storm database model.

See:
- https://laravel.com/docs/9.x/eloquent#pruning-models
- https://laravel.com/docs/9.x/eloquent#replicating-models

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -223,7 +223,7 @@ class Model extends EloquentModel implements ModelInterface
 
                 // No replicated / pruned events
                 if (
-                    ($radical === 'replicat' || $radical === 'prun') 
+                    ($radical === 'replicat' || $radical === 'prun')
                     && $event === 'ed'
                 ) {
                     continue;
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -213,13 +213,18 @@ class Model extends EloquentModel implements ModelInterface
             return;
         }
 
-        $radicals = ['creat', 'sav', 'updat', 'delet', 'fetch'];
+        $radicals = ['creat', 'sav', 'updat', 'delet', 'fetch', 'prun', 'replicat'];
         $hooks = ['before' => 'ing', 'after' => 'ed'];
 
         foreach ($radicals as $radical) {
             foreach ($hooks as $hook => $event) {
                 $eventMethod = $radical . $event; // saving / saved
                 $method = $hook . ucfirst($radical); // beforeSave / afterSave
+
+                if (($radical === 'replicat' || $radical === 'prun') 
+                     && $event === 'ed') {
+                    continue;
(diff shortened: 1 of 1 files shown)


LukeTowers:
@marvindurot are there replicated and pruned events too?

marvindurot:
> @marvindurot are there replicated and pruned events too?

No, these events don't exist, which is why I didn't add them.

src/Database/Model.php
```diff
@@ -213,14 +213,19 @@ protected function bootNicerEvents()
...
 
         foreach ($radicals as $radical) {
             foreach ($hooks as $hook => $event) {
                 $eventMethod = $radical . $event; // saving / saved
                 $method = $hook . ucfirst($radical); // beforeSave / afterSave
 
+                if (($radical === 'replicat' || $radical === 'prun') 
+                     && $event === 'ed') {
+                    continue;
+                }
```
```suggestion
                // No replicated / pruned events
                if (
                    ($radical === 'replicat' || $radical === 'prun') 
                    && $event === 'ed'
                ) {
                    continue;
                }
```

src/Database/Model.php
```diff
@@ -213,14 +213,22 @@ protected function bootNicerEvents()
...
         $hooks = ['before' => 'ing', 'after' => 'ed'];
 
         foreach ($radicals as $radical) {
             foreach ($hooks as $hook => $event) {
                 $eventMethod = $radical . $event; // saving / saved
                 $method = $hook . ucfirst($radical); // beforeSave / afterSave
 
+                // No replicated / pruned events
+                if (
+                    ($radical === 'replicat' || $radical === 'prun')
```
```suggestion
                    ($radical === 'replicat' || $radical === 'prun')
```

LukeTowers:
@bennothommo Could you look at this? I'm not seeing anything obvious that would be causing the tests to fail

bennothommo:
@LukeTowers @marvindurot the issue is that Laravel's base model doesn't have a `pruning` method. The method is provided by the `Prunable` trait.

This means that in our model, when it boots the "nicer" events and hits "beforePrune", the call to `pruning` is passed to the `__call` magic method which tries to feed it through to a scope method, or failing that, tries to pass it to an extension. When it tries to do it to the extensions, it's trying to load the default database configuration which doesn't exist on our test copy so that causes the test to fail.

We have three options:
- We can fix the test environment to have a default config (I would rather not do this, because we want the tests to run without a database needed)
- We can extend Laravel's `Prunable` trait to add our own events where needed
- We can add a `pruning` method to our base model (seems kinda dirty if it's not needed unless the `Prunable` trait is used)

LukeTowers:
@marvindurot do you have any thoughts on what @bennothommo said?

PR #151, opened 2023-08-09
Update tests for config writer v1.1.0

jaxwilko:
This should fix the test errors introduced with https://github.com/wintercms/laravel-config-writer/pull/2


mjauvin:
Can that be merged so we don't get those pesky errors?

PR #152, opened 2023-08-14
[PHP 8.3] Fix `get_class()` deprecations

Ayesh:
In PHP 8.3, [calling `get_class()` and `get_parent_class()` functions without arguments is deprecated](https://php.watch/versions/8.3/get_class-get_parent_class-parameterless-deprecated).

This fixes them with identical alternatives that do not cause the deprecation notice.

References:
 - [PHP RFC: Deprecate functions with overloaded signatures](https://wiki.php.net/rfc/deprecate_functions_with_overloaded_signatures)
 - [PHP 8.3: get_class() and get_parent_class() function calls without arguments deprecated](https://php.watch/versions/8.3/get_class-get_parent_class-parameterless-deprecated)

### DIFF
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -386,7 +386,7 @@ trait ExtendableTrait
             }
         }
 
-        $parent = get_parent_class();
+        $parent = get_parent_class($this);
         if ($parent !== false && method_exists($parent, '__get')) {
             return parent::__get($name);
         }
@@ -413,7 +413,7 @@ trait ExtendableTrait
         /*
          * This targets trait usage in particular
          */
-        $parent = get_parent_class();
+        $parent = get_parent_class($this);
         if ($parent !== false && method_exists($parent, '__set')) {
             parent::__set($name, $value);
         }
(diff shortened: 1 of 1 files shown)


bennothommo:
@Ayesh unfortunately that won't work - we have tried that before.

We'll have to refactor the `ExtendableTrait` to use the Reflection library in order to properly handle this case,  I believe.

LukeTowers:
@bennothommo why does it not work?

bennothommo:
@LukeTowers I think it's a quirk of traits - the `parent` keyword is (AFAIK) compiled on initialization, and inside the scope of a trait, it fails because a trait has no parent.

LukeTowers:
Should we log that as an issue with PHP 8.3 then?

bennothommo:
Evidentally, it's an issue with PHP 8.0-8.2 as well, going by our tests. I'm just about to roll out a PR that does what we need with Reflection classes. It should be way more reliable.

Ayesh:
I think the approach in #153 is more robust. However, I wanted to note that `get_parent_class($this)` should work even when they are in traits ([eval](https://3v4l.org/XMMcR)).

LukeTowers:
@Ayesh if that's the case are you able to figure out why the tests failed in our case with this PR? I would prefer the simpler approach of just passing $this if we can make that work, although I haven't dug into @bennothommo's PR too deeply yet.

PR #153, opened 2023-08-14
Remove deprecated "get_parent_class" calls in Extendable trait.

bennothommo:
This uses Reflection to instead determine the parent class and available magic methods to pass through to. In order to prevent infinite looping (which could've potentially been a problem before), it will also ignore any "extendable" classes when determining the parent.

Also fixed some tests that were using undefined class properties, and were also throwing deprecation errors.

Replaces https://github.com/wintercms/storm/pull/152

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -71,6 +71,11 @@ parameters:
 			path: src/Database/Model.php
 
 		-
+			message: "#^Static property Illuminate\\\\Database\\\\Eloquent\\\\Model\\:\\:\\$dispatcher \\(Illuminate\\\\Contracts\\\\Events\\\\Dispatcher\\) in isset\\(\\) is not nullable\\.$#"
+			count: 1
+			path: src/Database/Model.php
+
+		-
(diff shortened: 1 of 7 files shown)


mjauvin:
Can't we use `parent::class` instead? But as far as I can see, `get_parent_class($this)` is still supported.

LukeTowers:
Is there any performance differences using the reflection API instead of get_class_parent?

bennothommo:
@mjauvin I can tell you that so far, I have not been able to find a way to use any sort of `parent` call in this context that doesn't end up in an infinite loop or throwing the error that came up in the #152 test cases. If someone smarter than me is able to figure it out, by all means. 🙂

@LukeTowers I haven't timed it yet, but I'm going to assume there's a small performance hit. But as above, it may be the only way to do it if we want to allow the magic methods of parent classes to still be fallen back upon.

mjauvin:
@bennothommo and using `get_parent_class($this)` doesn't work either?

bennothommo:
@mjauvin It didn't when I tried it out, and it didn't in the test cases on the other PR I mentioned.

LukeTowers:
@bennothommo are you able to fix the code analysis error?

bennothommo:
@LukeTowers done

PR #155, opened 2023-08-31
Add path enumerable trait

bennothommo:
Implements path enumeration hierarchy for models, operating similar to nested set models but not requiring the rigid left/right values.

This is in preparation for some work towards database storage of media items.


src/Database/Traits/PathEnumerable.php
```diff
@@ -0,0 +1,322 @@
...
+     *
+     * This will take into account any parent changes, allowing you to get the new path before the record is saved.
+     */
+    public function getEnumerablePath(): string
+    {
+        if ($this->parent()->exists()) {
+            return $this->parent->{$this->getPathColumnName()} . '/' . $this->id;
+        }
+
+        return '/' . $this->id;
```
LukeTowers: @bennothommo Does this have to be hardcoded to reference the `id` column? What if I wanted to use a custom column name, say `folder` *cough cough*

bennothommo: We can make a constant for the heathens.

LukeTowers: What are the practical reasons for not wanting to use the actual folder names for the Media Library use case?

bennothommo: @LukeTowers probably nothing in terms of queries and functionality, but it could potentially result in larger path strings which would mean less levels of directories before you hit MySQL's string limits. At least with IDs you'll have a lot more levels before hitting those limits.

LukeTowers: Fair enough, do you think it would be worth adding in a constant for the max length of the path to surface that before it causes issues?

bennothommo: @LukeTowers I would say if the intention is to try and catch it before it becomes a problem, it's probably worth querying the schema directly and working out the maximum length, as opposed to an arbitrary max length constant. But either way, it will mean that the "segment" column (id/folder/etc.) would need to be configurable in some fashion, so yeah, it's either a constant or a method to get that column.

LukeTowers:
@bennothommo 👀

src/Database/Traits/PathEnumerable.php
bennothommo: @LukeTowers I've added the ability to define a custom segment column for the path.

bennothommo: Further to this, I opted not to try and find the maximum length of the path column in the schema as the only way to do so is to run an SQL query, which means it would call this every page load that a enumerable path model is used (unless it were cached).

I think this is something we can opt to document as opposed to trying to technically handle - most people, I imagine, would opt to use the ID paths, and for people who know how to change the segment column, they (should?) be knowledgeable enough to ensure their DB schema accounts for the potentially larger path values.

LukeTowers:
@bennothommo let me know when you want me to review this!

bennothommo:
@LukeTowers it's good to go :)

LukeTowers (approved):
LGTM @bennothommo, targeting next release pending docs PR for this feature :)

LukeTowers:
@bennothommo is there a docs PR for this?

bennothommo:
@LukeTowers not yet - I'm toying with the idea of unifying the tree traits in some fashion so they share a common base level of functionality and API.

LukeTowers:
@bennothommo is that something that needs to happen now or could we move forward with merging this so that we can get metadata support added to the Media Library?

bennothommo:
@LukeTowers my concern is that we'll have difficulty changing the API later.

LukeTowers:
@bennothommo status on this?

bennothommo:
Just waiting for docs. I mentioned previously trying to attempt to unify the tree APIs, but I probably won't do it for this branch. It might be worth looking into next major release.

LukeTowers:
@bennothommo sounds good to me, lets get the docs done up so we can roll this into the next release then. Would be nice to make some progress on the media library metadata.

LukeTowers:
@bennothommo ping

LukeTowers:
@bennothommo what specifically needs to be documented?

bennothommo:
Docs PR is here:  https://github.com/wintercms/docs/pull/227

PR #156, opened 2023-09-09
Improve automatic detaching / deletion of relations

mjauvin:
First step in Fixing https://github.com/wintercms/winter/issues/386

Related to https://github.com/wintercms/winter/pull/972

When deleting/soft-deleting & restoring a model relations, we need a separate case for relations using a pivot table.

### DIFF
--- a/src/Auth/Models/User.php
+++ b/src/Auth/Models/User.php
@@ -39,7 +39,7 @@ class User extends Model implements \Illuminate\Contracts\Auth\Authenticatable
      * @var array Relations
      */
     public $belongsToMany = [
-        'groups' => [Group::class, 'table' => 'users_groups']
+        'groups' => [Group::class, 'table' => 'users_groups', 'delete' => true],
     ];
 
     public $belongsTo = [
--- a/src/Auth/Models/User.php
+++ b/src/Auth/Models/User.php
@@ -39,7 +39,7 @@ class User extends Model implements \Illuminate\Contracts\Auth\Authenticatable
      * @var array Relations
      */
     public $belongsToMany = [
-        'groups' => [Group::class, 'table' => 'users_groups', 'delete' => true],
+        'groups' => [Group::class, 'table' => 'users_groups'],
     ];
 
     public $belongsTo = [
--- a/src/Auth/Models/User.php
+++ b/src/Auth/Models/User.php
@@ -154,17 +154,6 @@ class User extends Model implements \Illuminate\Contracts\Auth\Authenticatable
         $this->forceSave();
     }
 
-    /**
-     * Delete the user groups
-     * @return void
-     */
-    public function afterDelete()
-    {
-        if ($this->hasRelation('groups')) {
-            $this->groups()->detach();
-        }
-    }
-
     //
     // Persistence (used by Cookies and Sessions)
     //
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1013,6 +1013,9 @@ class Model extends EloquentModel implements ModelInterface
                 }
 
                 if (in_array($type, ['belongsToMany', 'morphToMany', 'morphedByMany'])) {
+                    if (!Arr::get($options, 'detach', true)) {
+                        continue;
+                    }
                     // we want to remove the pivot record, not the actual relation record
                     $relation()->detach();
                 } else {
@@ -1026,17 +1029,6 @@ class Model extends EloquentModel implements ModelInterface
                         $relation->each(function ($model) {
                             $model->forceDelete();
                         });
-                    }
-                }
-            }
-
-            /*
-             * Belongs-To-Many should clean up after itself always
-             */
-            if ($type == 'belongsToMany') {
-                foreach ($relations as $name => $options) {
-                    if (Arr::get($options, 'detach', true)) {
-                        $this->{$name}()->detach();
                     }
                 }
             }
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1018,10 +1018,11 @@ class Model extends EloquentModel implements ModelInterface
                     if (Arr::get($options, 'detach', true)) {
                         $this->{$name}()->detach();
                     }
-                } else if (in_array($type, ['hasOneThrough', 'hasManyThrough'])) {
-                    // the model does not own the relation, we should not remove it.
+                } else if (in_array($type, ['belongsTo', 'hasOneThrough', 'hasManyThrough', 'morphTo'])) {
+                    // the model does not own the related record, we should not remove it.
                     continue;
                 } else {
+                    // attachOne, attachMany, hasOne, hasMany, morphOne, morphMany
                     if (!Arr::get($options, 'delete', false)) {
                         continue;
                     }
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1021,8 +1021,7 @@ class Model extends EloquentModel implements ModelInterface
                 } elseif (in_array($type, ['belongsTo', 'hasOneThrough', 'hasManyThrough', 'morphTo'])) {
                     // the model does not own the related record, we should not remove it.
                     continue;
-                } else {
-                    // attachOne, attachMany, hasOne, hasMany, morphOne, morphMany
+                } elseif (in_array($type, ['attachOne', 'attachMany', 'hasOne', 'hasMany', 'morphOne', 'morphMany'])) {
                     if (!Arr::get($options, 'delete', false)) {
                         continue;
                     }
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1022,8 +1022,7 @@ class Model extends EloquentModel implements ModelInterface
 
                     if ($relation instanceof EloquentModel) {
                         $relation->forceDelete();
-                    }
-                    elseif ($relation instanceof CollectionBase) {
+                    } elseif ($relation instanceof CollectionBase) {
                         $relation->each(function ($model) {
                             $model->forceDelete();
                         });
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1018,7 +1018,7 @@ class Model extends EloquentModel implements ModelInterface
                     if (Arr::get($options, 'detach', true)) {
                         $this->{$name}()->detach();
                     }
-                } else if (in_array($type, ['belongsTo', 'hasOneThrough', 'hasManyThrough', 'morphTo'])) {
+                } elseif (in_array($type, ['belongsTo', 'hasOneThrough', 'hasManyThrough', 'morphTo'])) {
                     // the model does not own the related record, we should not remove it.
                     continue;
                 } else {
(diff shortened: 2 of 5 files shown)


mjauvin:
@xyzqtc @arvislacis welcome to join the discussion

mjauvin:
I tested this with success on a model with a `belongsToMany` relation, both deleting and restoring the model soft-deleted and restored the related models in the pivot table.

mjauvin:
@arvislacis do you mind creating a PR in the main winter repo to address the issue I mentioned in this PR description?

Thanks!

arvislacis:
@mjauvin I added https://github.com/wintercms/winter/pull/972 PR but it's not working for me - delete_at column is not being updated when deleting User and columns are fully deleted from `backend_users_groups` pilot table but maybe I am missing some critical part here. Feel free to comment on my PR.

mjauvin:
@arvislacis did you properly copy the SoftDelete trait from my PR to your local storm folder?

arvislacis:
@mjauvin I am pretty sure I did, I copied full RAW file of this PR to my local copy. But I will re-check this tomorrow.

mjauvin:
And how did you verify it's not working exactly ?

arvislacis:
@mjauvin Some updates on this:

1. Yes, I applied the patch correctly (to be clear);
2. For testing purposes I added break-point at the end of `updatePivotDeletedAtColumn` method - so far process executed as expected - the `deleted_at` column of pivot table was correctly updated with timestamp but if I removed the break-point then somewhere later in the process the relationship entry was still fully removed - I will try to investigate further.

"How did you verify it's not working exactly" - I verified it as described in https://github.com/wintercms/winter/issues/386#issuecomment-1384342881 - create new user or add existing user (non default admin) to group, then delete the user, then restore user with WinterCMS admin UI - result is that restored user entry has no group relationships.

arvislacis:
@mjauvin I did some further debugging but unfortunately didn't find root cause of pivot table relationship being still deleted  - it looks like that after the `softDelete` another delete operation is happening (maybe on `User` model itself which triggers pivot relationships to be still deleted).

Anyways, I would suggest to repeat user deletion and restoration operations with WinterCMS back-end as described here: https://github.com/wintercms/winter/issues/386#issuecomment-1384342881

mjauvin:
@arvislacis I found the problem, the Backend User model extends Storm's Auth/User model which has a afterDelete method shown below:
```php
    /**
     * Delete the user groups
     * @return void
     */
    public function afterDelete()
    {   
        if ($this->hasRelation('groups')) {
            $this->groups()->detach();
        }
    }
```

mjauvin:
~~So just override that method as an empty method in Winter's model and that resolves the issue.~~

~~In your PR, add a comment that explains why the method is overriden and empty.~~

mjauvin:
Actually, we might want to remove that `afterDelete` method in Storm's model since it's already done by default in `Database/Model`...

@LukeTowers or @bennothommo can you confirm that we should remove the afterDelete method in `Storm\Auth\Models\User` ?

arvislacis:
@mjauvin Nice find! I can confirm that applying the latest changes fixes the https://github.com/wintercms/winter/issues/386

For me it looks like `afterDelete` can be safely removed ad now we are soft deleting group relationships.

mjauvin:
Yeah, I made a fix to the base Model class so that the afterDelete method is not required anymore

src/Auth/Models/User.php
```diff
@@ -154,17 +154,6 @@ public function afterLogin()
...
-    /**
-     * Delete the user groups
-     * @return void
-     */
-    public function afterDelete()
-    {
-        if ($this->hasRelation('groups')) {
-            $this->groups()->detach();
-        }
-    }
```
bennothommo, line 166: @mjauvin this won't have any ramifications with the User plugin, yeah? Since that also extends this class (and potentially other plugins do too).

src/Database/Model.php
```diff
@@ -1008,21 +1008,26 @@ protected function performDeleteOnRelations()
...
+                    $relation()->detach();
+                } else {
+                    if (!Arr::get($options, 'delete', false)) {
+                        continue;
+                    }
+
+                    if ($relation instanceof EloquentModel) {
+                        $relation->forceDelete();
+                    }
+                    elseif ($relation instanceof CollectionBase) {
```
bennothommo:
```suggestion
                    } elseif ($relation instanceof CollectionBase) {
```

src/Database/Traits/SoftDelete.php
```diff
@@ -177,6 +183,21 @@ public function restore()
...
+    protected function updatePivotDeletedAtColumn(string $relationName, array $options, string|null $value)
+    {
+        $relation = $this->{$relationName}();
+
+        // get deletedAtColumn from the relation options, otherwise use default
+        $deletedAtColumn = array_get($options, 'deletedAtColumn', 'deleted_at');
+
+        $query = $relation->newPivotQuery()->update([
+            $deletedAtColumn => $value,
+        ]);
```
bennothommo: @mjauvin best to use the [`getDeletedAtColumn()` method](https://github.com/wintercms/storm/blob/develop/src/Database/Traits/SoftDelete.php#L267-L275) here:

```suggestion
        $relation->newPivotQuery()->update([
            $relation->getDeletedAtColumn() => $value,
        ]);
```

bennothommo (dismissed):

src/Database/Traits/SoftDelete.php
```diff
@@ -123,13 +123,19 @@ protected function performSoftDeleteOnRelations()
...
-                    });
+                if (in_array($type, ['belongsToMany', 'morphToMany', 'morphedByMany'])) {
+                    // relations using pivot table
+                    $value = $this->fromDateTime($this->freshTimestamp());
+                    $this->updatePivotDeletedAtColumn($name, $options, $value);
+                } else {
+                    if ($relation instanceof EloquentModel) {
+                        $relation->delete();
+                    }
+                    elseif ($relation instanceof CollectionBase) {
```
bennothommo:
```suggestion
                    } elseif ($relation instanceof CollectionBase) {
```

src/Database/Traits/SoftDelete.php
```diff
@@ -191,18 +212,23 @@ protected function performRestoreOnRelations()
...
+                } else {
+                    $relation = $this->{$name}()->onlyTrashed()->getResults();
+                    if (!$relation) {
+                        continue;
+                    }
+
+                    if ($relation instanceof EloquentModel) {
+                        $relation->restore();
+                    }
+                    elseif ($relation instanceof CollectionBase) {
```
bennothommo:
```suggestion
                    } elseif ($relation instanceof CollectionBase) {
```

src/Database/Traits/SoftDelete.php
mjauvin: You can't really use the getDeletedAtColumn method as the relation is NOT using the softDelete trait

bennothommo: Ah of course, good point.

src/Auth/Models/User.php
mjauvin: ~~We may need to add 'delete' => true to the User plugin User belongsToMany groups relation, you're right~~

mjauvin: @bennothommo actually, this is fine as the pivot records for the relations now get deleted always as they should from the base Model class, unless of course the relation has been marked as "softDelete => true", in which case, only the deleted_at column would get updated.

So the previous behavior is preserved, the only difference is it's now done in the base Model's `performDeleteOnRelations()` method for all models instead of being "patched" in only for the Auth\User model.

LukeTowers:
@mjauvin is there a pending docs PR?

mjauvin:
> @mjauvin is there a pending docs PR?

Not yet  no.

mjauvin:
@LukeTowers https://github.com/wintercms/docs/pull/159

bennothommo:
@mjauvin could we get some unit tests for hard deletes and the different relation types, if it's not too much bother? Given this is deletion functionality with a potential for data loss if we're not careful, we should cover our bases.

src/Database/Model.php
```diff
@@ -1018,6 +1018,9 @@ protected function performDeleteOnRelations()
                     if (Arr::get($options, 'detach', true)) {
                         $this->{$name}()->detach();
                     }
+                } else if (in_array($type, ['hasOneThrough', 'hasManyThrough'])) {
```
```suggestion
                } elseif (in_array($type, ['hasOneThrough', 'hasManyThrough'])) {
```

src/Database/Model.php
```diff
@@ -1008,32 +1009,31 @@ protected function performDeleteOnRelations()
...
-             * Belongs-To-Many should clean up after itself always
-             */
-            if ($type == 'belongsToMany') {
-                foreach ($relations as $name => $options) {
+                if (in_array($type, ['belongsToMany', 'morphToMany', 'morphedByMany'])) {
+                    // we want to remove the pivot record, not the actual relation record
                     if (Arr::get($options, 'detach', true)) {
                         $this->{$name}()->detach();
                     }
+                } else if (in_array($type, ['belongsTo', 'hasOneThrough', 'hasManyThrough', 'morphTo'])) {
```
```suggestion
                } elseif (in_array($type, ['belongsTo', 'hasOneThrough', 'hasManyThrough', 'morphTo'])) {
```

src/Database/Model.php
```diff
@@ -1008,32 +1009,31 @@ protected function performDeleteOnRelations()
...
-                foreach ($relations as $name => $options) {
+                if (in_array($type, ['belongsToMany', 'morphToMany', 'morphedByMany'])) {
+                    // we want to remove the pivot record, not the actual relation record
                     if (Arr::get($options, 'detach', true)) {
                         $this->{$name}()->detach();
                     }
+                } elseif (in_array($type, ['belongsTo', 'hasOneThrough', 'hasManyThrough', 'morphTo'])) {
+                    // the model does not own the related record, we should not remove it.
+                    continue;
+                } else {
```
LukeTowers: Shouldn't we be explicit here like with the other checks to ensure that any future relationship types that may be added don't get accidentally processed here?

mjauvin: Yeah, I thought about it and decided not to. Will make the change.

LukeTowers:
LGTM! @bennothommo any final thoughts?

mjauvin:
@bennothommo I added the hard delete tests you requested.

bennothommo (approved):

arvislacis:
@LukeTowers @mjauvin As this is merged now, what about related PR and issue in winter repo?

mjauvin:
What do you mean exactly?

arvislacis:
I mean merging also https://github.com/wintercms/winter/pull/972 and closing https://github.com/wintercms/winter/issues/386 or it will be done with WinterCMS 1.4 release?

mjauvin:
Yes, theee should be merged and closed soon.

PR #159, opened 2023-10-18
Restore previous tests order

mjauvin:

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1022,6 +1022,7 @@ class Model extends EloquentModel implements ModelInterface
                         continue;
                     }
 
+                    // Attempt to load the related record(s)
                     if (!$relation = $this->{$name}) {
                         continue;
                     }
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1009,10 +1009,6 @@ class Model extends EloquentModel implements ModelInterface
              * Hard 'delete' definition
              */
             foreach ($relations as $name => $options) {
-                if (!$relation = $this->{$name}) {
-                    continue;
-                }
-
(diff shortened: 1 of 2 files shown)


mjauvin:
I couldn't reproduce the errors while calling $model->{'relation'} for all relation types with REAL models... I am going to assume the test models are missing something and merge this fix.

LukeTowers:
What is that conditional supposed to be checking for?

mjauvin:
> What is that conditional supposed to be checking for?

Not entirely sure, to be honest.

LukeTowers:
> Not entirely sure, to be honest.

I don't really like that answer 😜 time to go digging through the history I guess 😂

src/Database/Model.php
```diff
@@ -1026,6 +1022,10 @@ protected function performDeleteOnRelations()
                         continue;
                     }
 
+                    if (!$relation = $this->{$name}) {
```
LukeTowers, line 1026:
```suggestion
                    // Attempt to load the related record(s)
                    if (!$relation = $this->{$name}) {
```

src/Database/Traits/SoftDelete.php
```diff
@@ -115,10 +115,10 @@ protected function performSoftDeleteOnRelations()
         $definitions = $this->getRelationDefinitions();
         foreach ($definitions as $type => $relations) {
             foreach ($relations as $name => $options) {
-                if (!$relation = $this->{$name}) {
+                if (!array_get($options, 'softDelete', false)) {
                     continue;
                 }
-                if (!array_get($options, 'softDelete', false)) {
+                if (!$relation = $this->{$name}) {
```
LukeTowers, line 122:
```suggestion
                // Attempt to load the related record(s)
                if (!$relation = $this->{$name}) {
```

PR #165, opened 2024-01-29
Fix dynamic properties on PHP 8.2+

mjauvin:
PHP 8.2+ has deprecated dynamic properties. This PR addresses the issue by using the extensionData['dynamicProperties'] array for the dynamic properties.

### DIFF
--- a/src/Halcyon/Model.php
+++ b/src/Halcyon/Model.php
@@ -1603,6 +1603,15 @@ class Model extends Extendable implements ModelInterface, ArrayAccess, Arrayable
         static::$mutatorCache[$class] = $mutatedAttributes;
     }
 
+    public function addDynamicProperty($dynamicName, $value = null)
+    {
+        if (isset($this->attributes[$dynamicName])) {
+            return;
+        }
+
+        $this->setAttribute($dynamicName, $value);
+    }
+
     /**
      * Dynamically retrieve attributes on the model.
      *
--- a/src/Halcyon/Model.php
+++ b/src/Halcyon/Model.php
@@ -1608,10 +1608,10 @@ class Model extends Extendable implements ModelInterface, ArrayAccess, Arrayable
         if (isset($this->attributes[$dynamicName])) {
             return;
         }
-
-        parent::addDynamicProperty($dynamicName, $value);
-
         $this->setAttribute($dynamicName, $value);
+
+        // only need to be set to exclude those attributes from getSettingsAttribute()
+        array_set($this->extensionData['dynamicProperties'], $dynamicName, null);
     }
 
     /**
--- a/src/Halcyon/Model.php
+++ b/src/Halcyon/Model.php
@@ -1609,6 +1609,8 @@ class Model extends Extendable implements ModelInterface, ArrayAccess, Arrayable
             return;
         }
 
+        parent::addDynamicProperty($dynamicName, $value);
+
         $this->setAttribute($dynamicName, $value);
     }
 
--- a/src/Halcyon/Model.php
+++ b/src/Halcyon/Model.php
@@ -1603,17 +1603,6 @@ class Model extends Extendable implements ModelInterface, ArrayAccess, Arrayable
         static::$mutatorCache[$class] = $mutatedAttributes;
     }
 
-    public function addDynamicProperty($dynamicName, $value = null)
-    {
-        if (isset($this->attributes[$dynamicName])) {
-            return;
-        }
-        $this->setAttribute($dynamicName, $value);
-
-        // only need to be set to exclude those attributes from getSettingsAttribute()
-        array_set($this->extensionData['dynamicProperties'], $dynamicName, null);
-    }
-
     /**
      * Dynamically retrieve attributes on the model.
      *
--- a/src/Halcyon/Model.php
+++ b/src/Halcyon/Model.php
@@ -1611,7 +1611,7 @@ class Model extends Extendable implements ModelInterface, ArrayAccess, Arrayable
      */
     public function __get($key)
     {
-        return $this->getAttribute($key);
+        return $this->extendableGet($key) ?? $this->getAttribute($key);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


mjauvin:
Still need to add the unit tests, but looks like this is working beautifully.

src/Halcyon/Model.php
```diff
@@ -1603,6 +1603,17 @@ public static function cacheMutatedAttributes($class)
...
+    {
+        if (isset($this->attributes[$dynamicName])) {
+            return;
+        }
+        $this->setAttribute($dynamicName, $value);
+
+        // only need to be set to exclude those attributes from getSettingsAttribute()
+        array_set($this->extensionData['dynamicProperties'], $dynamicName, null);
+    }
+
```
bennothommo: @mjauvin I'm not sure I'm a fan of dynamic properties being populated into the model attributes. The key purpose of that method is to populate class properties. Is there a scenario where attribute modification would be needed?

mjauvin: Not a fan either, I was trying to keep the existing unit tests happy, and one of those checks if the attribute exists after adding a dynamic property.

LukeTowers: Is there any information in the git blame for that existing test as to why it was a test or is it just a case of testing the existing behaviour as it was because that's what PHP level dynamic properties would have done to the Model class previously?

mjauvin: Looks like I'm the one to blame for that... :rofl: 

```
b8ca7aa51 tests/Halcyon/ModelTest.php        (Marc Jauvin    2018-12-18 11:23:46 -0500 356)         $page = HalcyonTestPage::create([
b8ca7aa51 tests/Halcyon/ModelTest.php        (Marc Jauvin    2018-12-18 11:23:46 -0500 357)             'fileName' => 'dynamicproperty',
b8ca7aa51 tests/Halcyon/ModelTest.php        (Marc Jauvin    2018-12-18 11:23:46 -0500 358)             'title' => 'Add Dynamic Property',
b8ca7aa51 tests/Halcyon/ModelTest.php        (Marc Jauvin    2018-12-18 11:23:46 -0500 359)             'markup' => '<p>Foo bar!</p>'
b8ca7aa51 tests/Halcyon/ModelTest.php        (Marc Jauvin    2018-12-18 11:23:46 -0500 360)         ]);
b8ca7aa51 tests/Halcyon/ModelTest.php        (Marc Jauvin    2018-12-18 11:23:46 -0500 361) 
b8ca7aa51 tests/Halcyon/ModelTest.php        (Marc Jauvin    2018-12-18 11:23:46 -0500 362)         $page->addDynamicProperty('myDynamicProperty', 'myDynamicPropertyValue');
e4570b738 tests/Halcyon/ModelTest.php        (Marc Jauvin    2019-06-04 10:04:44 -0400 363)         $this->assertEquals('myDynamicPropertyValue', $page->myDynamicProperty);
```

mjauvin: The question now is what do we do with dynamic class properties for the Halcyon model? Without adding this to the attributes, there is no way to access (get/set) those properties once they've been added since Halcyon overrides the __get & __set methods to work with the attributes array.

mjauvin: The Database/Model class also adds an `addDynamicProperty()` method of its own (adds the property to the purgeable array).

mjauvin: We need the dynamic properties, specifically in `Winter.Translate` plugin where it adds `TranslatablePage` behavior to the Cms\Classes\Page Halcyon model.

LukeTowers: Can we add in logic in the overrides to retrieve from the dynamic properties if set over / above the attributes container?

Leave the __set() to treat them as attributes if done manually, but change the __get() to check explicit dynamic properties first. Any thoughts @bennothommo @mjauvin @jaxwilko?

mjauvin: Yeah, I started doing this but then dynamic properties cannot be updated. Kind of flaky.

bennothommo: @mjauvin the test itself might be incorrect.

The dynamic property isn't being populated to the attributes array (which is correct - it isn't an attribute brought in by Halcyon), but it would still be expected to be a property of the class, which means if you were to dynamically populate the `translatable` property on the class, the behaviour should still be able to pick it up if the `__get` magic method looks in the dynamic properties array first.

The only issue I see is potentially people are using `addDynamicProperty()` to populate an attribute into the model for it to be saved, since this method did set properties directly on the model previously (which would've been picked up by the `__set` magic method in the Halcyon model and added to its attributes array). I would argue this is incorrect usage - the method name `addDynamicProperty()` hints that it is "dynamic" - it's value is being populated by a third-party and could change at any time.

In this instance, I would expect that we would need to inform people that in these instances, they should perhaps be using events instead.

mjauvin: Ok, I updated the PR with this, works as expected.

mjauvin:
@bennothommo any idea where the errors in `Code Analysis` come from ?

bennothommo:
Nope - try merging in the develop branch maybe?

mjauvin:
It already is.

LukeTowers:
Is there anything left on this @mjauvin @bennothommo or is it ready for review?

mjauvin:
It's ready for review

mjauvin:
@bennothommo @LukeTowers do you know if this line here necessarily refers to a real class property or if it could be a dynamic property?

https://github.com/wintercms/storm/blob/develop/src/Extension/ExtendableTrait.php#L383

bennothommo:
@mjauvin that's looking for a real property on the actual extension class (ie. the behaviour). It allows the public properties of the behaviour to be included in the extended class (like what would happen with a trait).

mjauvin:
> @mjauvin that's looking for a real property on the actual extension class (ie. the behaviour). It allows the public properties of the behaviour to be included in the extended class (like what would happen with a trait).

Ok, so I don't need to convert this to use `$extensionObject->propertyExists()` to also account for dynamic properties that would have been added to the behavior since currently behaviors are not extendable.

Thanks for confirming this.

PR #166, opened 2024-02-16
Make the dynamic property writable again

mjauvin:

### DIFF
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -403,10 +403,13 @@ trait ExtendableTrait
         $parent = $this->extensionGetParentClass();
         if ($parent !== false && $this->extensionMethodExists($parent, '__set')) {
             $this->extensionCallMethod($parent, '__set', [$name, $value]);
+            return;
         }
 
         // local dynamic property
-        array_set($this->extensionData['dynamicProperties'], $name, $value);
+        if (array_key_exists($name, $this->getDynamicProperties())) {
+            array_set($this->extensionData['dynamicProperties'], $name, $value);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mjauvin:
When you get a chance, can you look at the phpstan error @bennothommo ?

src/Extension/ExtendableTrait.php
```diff
@@ -412,12 +403,11 @@ public function extendableSet($name, $value)
...
             $this->extensionCallMethod($parent, '__set', [$name, $value]);
+            return;
         }
 
-        /*
-         * Setting an undefined property
-         */
-        if (!self::$extendableGuardProperties) {
+        // local dynamic property
+        if (array_key_exists($name, $this->getDynamicProperties())) {
```
bennothommo, line 411: @mjauvin what's this condition needed for?

bennothommo:
@mjauvin I re-ran the test and it passed. No idea why it failed before :man_shrugging:

src/Extension/ExtendableTrait.php
mjauvin: It prevents creating a dynamic property just by setting it (if there's no parent `__set` method for example)

mjauvin:
Ok, added a new unit test to be safe and this now works as expected.

Winter.Translate has been fixed to work properly with new dynamic properties as well.

src/Extension/ExtendableTrait.php
```diff
@@ -412,12 +403,11 @@ public function extendableSet($name, $value)
...
         if ($parent !== false && $this->extensionMethodExists($parent, '__set')) {
             $this->extensionCallMethod($parent, '__set', [$name, $value]);
+            return;
         }
 
-        /*
-         * Setting an undefined property
-         */
-        if (!self::$extendableGuardProperties) {
+        // local dynamic property
```
```suggestion
        // Don't allow automatic creation of dynamic properties through the setter magic method,
        // addDynamicProperty() must be used instead.
```

PR #167, opened 2024-02-19
Add support for Laravel-style relations

bennothommo:
This PR allows people to define Laravel-style relations using methods that return `Relation` objects. So instead of:

```php
public $belongsTo = [
     'user' => User::class,
];
```

You may use:

```php
public function user(): BelongsTo
{
    return $this->belongsTo(User::class);
}
```

The main reason for this is to reduce friction for porting Laravel code over to Winter, and as a preference point for people who may wish to use the more declarative style of defining relations.

I've adjusted the unit tests to demonstrate this in action, and will be porting over the relation tests in the Winter `system` module into Storm, as they really should be in Storm anyway.

In order to maintain parity with the array-style configuration of relations in Winter, method-defined relations will need to either implement one of the following to indicate that they provide relations, and allow certain functionality to find them as required (for example, deleting related models when deleting a primary model):

- The relation method must have a return type that correlates to a `Relation` class (like with the example above)
- Alternatively, the method must use the `Winter\Storm\Database\Attributes\Relation` attribute and define the relation type, like so:

```php
use Winter\Storm\Database\Attributes\Relation;

#[Relation('belongsTo')]
public function user()
{
    return $this->belongsTo(User::class);
}
```

### Todo

- [x] Fix attachment relations and tests
- [x] Add support for model traits with new format
- [x] Finish porting tests from Database Tester plugin in core
- [x] Full documentation of new syntax


bennothommo:
This PR is now ready for review. Will submit a PR to the docs that will indicate the API changes.

mjauvin:
Wow, that is quite extensive... Nice work!

How confident are you in the unit tests? ;)

bennothommo:
@mjauvin I've basically added copies of nearly every test we had to directly test the Laravel-style relations, so pretty confident :)

mjauvin:
So you're confident about the winter style relations as well? There were a LOT of changes...

bennothommo:
@mjauvin the original tests for the array-style relations are still there :)

bennothommo:
After giving this some thought, this might need to be released on a new "major" version update since we're not using semver.

The ramification of using this functionality - especially inside of a plugin - would lock that to the version of Winter that introduced it. So while it doesn't break backwards compatibility of the public API of Winter as it is now (ie. it doesn't cause any current functionality to fail), as soon as someone uses it, it immediately makes their Winter CMS install or plugin incompatible with older versions of Winter.

LukeTowers:
@bennothommo I'm fine with that, it'll provide some motivation to make the jump to Laravel 10 or 11 😉

LukeTowers:
@bennothommo I think I don't care about plugins choosing to become incompatible with older versions of Winter. Plugins are free to use semver and can define their requirements of a specific winter version in their composer.json; so I don't think this needs to wait for a new major version.

What is needed to get this across the line?

bennothommo:
@LukeTowers we'd just need to fix the merge conflict and tests. Otherwise, it's functionally complete.

src/Database/Attributes/Relation.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Winter\Storm\Database\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_METHOD)]
+class Relation
+{
+}
```
LukeTowers, line 55: What is this used for?

src/Database/Relations/Concerns/DefinedConstraints.php
```diff
@@ -130,4 +117,16 @@ public function addDefinedConstraintsToQuery($query, ?array $args = null)
             $query->$scope($this->parent);
         }
     }
+
+    /**
+     * Get the relation definition for the related model.
+     *
+     * @return array
```
LukeTowers:
```suggestion
```

src/Database/Relations/Concerns/DefinedConstraints.php
```diff
@@ -11,20 +14,18 @@ trait DefinedConstraints
     /**
      * Set the defined constraints on the relation query.
      *
+     * This method is kept for backwards compatibility, but is no longer being called directly by Storm when
+     * initializing relations. Constraints are now applied on an as-needed basis.
```
LukeTowers: Where are they being applied on the as-needed basis? Also should we tag this as deprecated if we're not going to use it anymore?

src/Database/Relations/Concerns/DefinedConstraints.php
```diff
@@ -60,26 +64,6 @@ public function addDefinedConstraintsToRelation($relation, ?array $args = null)
...
-            } else {
-                $foreignKey = $relation->getForeignKey();
-            }
-            $countSql = $this->parent->getConnection()->raw('count(*) as count');
-            $relation
-                ->select($foreignKey, $countSql)
-                ->groupBy($foreignKey)
-                ->orderBy($foreignKey)
-            ;
-        }
```
LukeTowers, line 82: Where is this handled now?

src/Database/Relations/Concerns/CanBePushed.php
```diff
@@ -0,0 +1,69 @@
...
+ * {
+ *     return $this->hasMany(Message::class)->push();
+ * }
+ * ```
+ *
+ * If you are using the array-style definition, you can use the `push` key to mark the relationship as pushable or not.
+ *
+ * ```php
+ * public $hasMany = [
+ *     'messages' => [Message::class, 'push' => true]
```
LukeTowers, line 26:
```suggestion
 *     'messages' => [Message::class, 'push' => false]
```

src/Database/Relations/Concerns/CanBePushed.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+namespace Winter\Storm\Database\Relations\Concerns;
+
+/**
+ * This trait is used to mark certain relationships as pushable.
```
LukeTowers, line 6: Is this something new from Laravel?

src/Database/Relations/Concerns/CanBeExtended.php
```diff
@@ -0,0 +1,126 @@
...
+
+namespace Winter\Storm\Database\Relations\Concerns;
+
+use Winter\Storm\Extension\ExtendableTrait;
+
+/**
+ * Adds extension support for relations.
+ *
+ * Since Laravel allows macros for relations, we should also provide Winter's extension framework for relations, to
+ * allow scenarios like model traits being able to define custom relation methods for configuring the relation.
```
LukeTowers, line 11: Could you add some examples here for what this change enables in Winter now, perhaps alongside the Laravel macros support? Trying to wrap my head around it.

src/Database/Relations/Concerns/CanBeDetachable.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+namespace Winter\Storm\Database\Relations\Concerns;
+
+/**
+ * This trait is used to mark certain relationships as detachable.
```
LukeTowers, line 6: Is this something new from Laravel?

src/Database/Relations/Concerns/CanBeDependent.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+namespace Winter\Storm\Database\Relations\Concerns;
+
+/**
+ * This trait is used to mark certain relationships as dependent.
```
LukeTowers, line 6: This one already existed in Winter under the `delete` option right?

src/Database/Relations/Concerns/CanBeCounted.php
```diff
@@ -0,0 +1,90 @@
...
+ * A relationship counter simply returns the number of related records as an integer. For backwards compatibility with
+ * the array-style definition, this is returned as an instance of the model, with only a `count` column provided.
+ *
+ * This trait can be used on all relationship types.
+ *
+ * This trait provides a declarative way to mark a relationship as a counter only, by simply adding `->countOnly()` to
+ * the relationship definition method. For example:
+ *
+ * ```php
+ * public function messages()
```
LukeTowers:
```suggestion
 * public function totalMessages()
```

src/Database/Relations/Concerns/CanBeCounted.php
```diff
@@ -0,0 +1,90 @@
...
+ * {
+ *     return $this->hasMany(Message::class)->countOnly();
+ * }
+ * ```
+ *
+ * If you are using the array-style definition, you can use the `count` key to mark the relationship as a counter only.
+ *
+ * ```php
+ * public $hasMany = [
+ *     'messages' => [Message::class, 'count' => true]
```
LukeTowers:
```suggestion
 *     'total_messages' => [Message::class, 'count' => true]
```

src/Database/Relations/Concerns/AttachOneOrMany.php
```diff
@@ -12,14 +12,14 @@ trait AttachOneOrMany
...
      */
-    protected $relationName;
+    protected $public;
 
     /**
-     * @var ?boolean Default value for file public or protected state.
+     * The field name (relation) to associate this attachment with.
      */
-    protected $public;
+    protected string $fieldName;
```
LukeTowers, line 22: Why was the order of these properties switched? Messier diff this way. Also what exactly is the difference between relationName and fieldName?

src/Database/Relations/BelongsToMany.php
```diff
@@ -1,10 +1,147 @@
...
+use Winter\Storm\Database\Pivot;
 
-class BelongsToMany extends BelongsToManyBase
+class BelongsToMany extends BelongsToManyBase implements Relation
 {
     use Concerns\BelongsOrMorphsToMany;
+    use Concerns\CanBeCounted;
+    use Concerns\CanBeDetachable;
+    use Concerns\CanBeExtended;
+    use Concerns\CanBePushed;
```
LukeTowers, line 17: Thoughts on the `CanBeX` vs `Xable` naming scheme?

src/Database/Relations/BelongsToMany.php
```diff
@@ -1,10 +1,147 @@
...
+            return;
+        }
+
+        /*
+         * Convert models to keys
+         */
+        if ($value instanceof Model) {
+            $value = $value->getKey();
+        }
+        elseif (is_array($value)) {
```
LukeTowers:
```suggestion
        } elseif (is_array($value)) {
```
We should probably make some adjustments to our published coding style and apply it in bulk across our codebases soon.

src/Database/Relations/BelongsToMany.php
```diff
@@ -1,10 +1,147 @@
...
+            foreach ($value as $_key => $_value) {
+                if ($_value instanceof Model) {
+                    $value[$_key] = $_value->getKey();
+                }
+            }
+        }
+
+        /*
+         * Convert scalar to array
+         */
```
LukeTowers:
```suggestion
        // Convert scalar to array
```

src/Database/Relations/BelongsToMany.php
```diff
@@ -1,10 +1,147 @@
...
+        if (!is_array($value) && !$value instanceof Collection) {
+            $value = [$value];
+        }
+
+        /*
+         * Setting the relationship
+         */
```
LukeTowers:
```suggestion
        // Setting the relationship
```

src/Database/Relations/BelongsToMany.php
```diff
@@ -1,10 +1,147 @@
...
+            // Perform sync when the model is saved
+            $this->parent->bindEventOnce('model.afterSave', function () {
+                $this->detach();
+            });
+            return;
+        }
+
+        /*
+         * Convert models to keys
+         */
```
LukeTowers:
```suggestion
        // Convert models to keys
```

src/Database/Relations/BelongsToMany.php
```diff
@@ -1,10 +1,147 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function setSimpleValue($value): void
+    {
+        $relationModel = $this->getRelated();
+
+        /*
+         * Nulling the relationship
+         */
```
LukeTowers:
```suggestion
        // Nulling the relationship
```

src/Database/Concerns/HasRelationships.php
```diff
@@ -279,520 +338,573 @@ protected function getRelationDefaults($type)
...
+        // Add relation name
+        $relation->setRelationName($relationName);
+
+        // Add dependency, if required
+        if (
+            in_array(
+                \Winter\Storm\Database\Relations\Concerns\CanBeDependent::class,
+                class_uses_recursive($relation)
+            )
+            && (($definition['delete'] ?? false) === true)
```
LukeTowers: Should we check the definition configs before making calls to reflection?

src/Database/Concerns/HasRelationships.php
```diff
@@ -279,520 +338,573 @@ protected function getRelationDefaults($type)
...
     }
 
     /**
-     * Define a one-to-one relationship.
-     * This code is a duplicate of Eloquent but uses a Storm relation class.
-     * @return \Winter\Storm\Database\Relations\HasOne
+     * Finds the calling function name from the stack trace.
      */
-    public function hasOne($related, $primaryKey = null, $localKey = null, $relationName = null)
+    protected function getRelationCaller(): ?string
```
LukeTowers, line 630: What is this being used for?

src/Database/Concerns/HasRelationships.php
```diff
@@ -279,520 +338,573 @@ protected function getRelationDefaults($type)
...
     }
 
     /**
-     * Define a polymorphic many-to-many inverse relationship.
-     * This code is almost a duplicate of Eloquent but uses a Storm relation class.
-     * @return \Winter\Storm\Database\Relations\MorphToMany
+     * {@inheritDoc}
      */
-    public function morphedByMany($related, $name, $table = null, $primaryKey = null, $foreignKey = null, $parentKey = null, $relatedKey = null, $relationName = null)
+    protected function newMorphMany(Builder $query, Model $parent, $type, $id, $localKey)
```
LukeTowers, line 1016: Are these `new$relationType` methods in the same order as the old ones? Just wondering in case we could get a cleaner diff for them. Also is there any concerns with the public methods being removed?

src/Database/Model.php
```diff
@@ -849,12 +849,10 @@ public function newPivot(EloquentModel $parent, array $attributes, $table, $exis
...
     {
-        $definition = $this->getRelationDefinition($relationName);
+        $relation = $this->{$relationName}();
+        $pivotModel = $relation->getPivotClass();
 
-        if (!is_null($definition) && array_key_exists('pivotModel', $definition)) {
-            $pivotModel = $definition['pivotModel'];
-            return $pivotModel::fromAttributes($parent, $attributes, $table, $exists);
-        }
+        return $pivotModel::fromAttributes($parent, $attributes, $table, $exists);
```
```suggestion
        return $pivotModel::fromRawAttributes($parent, $attributes, $table, $exists);
```

See https://github.com/wintercms/storm/commit/339af87267d355a8acef6dafbed9d2f2523099d2

src/Database/Relations/BelongsToMany.php
```diff
@@ -1,10 +1,147 @@
...
+        $value = [];
+
+        $sessionKey = $this->parent->sessionKey;
+
+        if ($this->parent->relationLoaded($this->relationName)) {
+            $related = $this->getRelated();
+
+            $value = $this->parent->getRelation($this->relationName)->pluck($related->getKeyName())->all();
+        }
+        else {
```
LukeTowers:
```suggestion
        } else {
```

LukeTowers (changes requested):

src/Database/Attributes/Relation.php
bennothommo: This attribute can be used instead of a Relation class return type to indicate a method is a relation method. It's to aid discovery of relations in a model, since some functionality of Winter requires knowledge of the relationship configuration without loading the relation directly (such as the soft delete trait needing to know relations of the related models in order to cascade soft deletions).

It could be used in cases where a relation is dynamic or if, for some reason, the dev doesn't want to use return types.

src/Database/Relations/Concerns/DefinedConstraints.php
bennothommo: This method used to pull double functionality. It set constraints on the relation object, and set constraints on the query, but this method called both on the relation object itself when retrieving the relation.

Setting constraints on the relation now is only needed when using property style relations, because in method style relations, you would add the constraints directly in the method, such as conditions, ordering, default models, timestamps, etc.

The query constraints are also only automatically applied on property style relations, and only when needed in the `Winter\Storm\Database\Concerns\HasRelationships` concern. I also saw it being used in the Relation Controller, so I kept that method available.

Thus, we no longer need this method - however, I couldn't discount that it might be used in a plugin somewhere.

bennothommo: Oh, and yes, we should mark it deprecated

src/Database/Relations/Concerns/DefinedConstraints.php
bennothommo: In the `Winter\Storm\Database\Relations\Concerns\CanBeCounted` concern. :)

src/Database/Relations/Concerns/CanBeDependent.php
bennothommo: That's right, but I didn't like the term deleteable or canDelete. It didn't quite convey what the functionality does.

src/Database/Relations/Concerns/CanBePushed.php
bennothommo: This example is to show how to enable push when using the property style, so it would need to be true.

src/Database/Relations/Concerns/CanBePushed.php
bennothommo: I believe it's been in Laravel for quite a while, and it's something that we support in Winter through the `push` option (see [our docs](https://wintercms.com/docs/v1.2/docs/database/relations#detailed-definitions) for more detail).

src/Database/Relations/Concerns/CanBeExtended.php
bennothommo: An example is for the `SoftDelete` trait - https://github.com/wintercms/storm/blob/c4abef88cb1c9909395e3d8e4cc176775e08d35b/src/Database/Traits/SoftDelete.php#L79-L112

I needed to allow for relations to be able to add chainable methods and additional properties for relation method definitions when circumstances require it, and Laravel macros don't allow you to define properties, only methods. In the case here with the `SoftDelete` trait, it is dynamically creating the "flag" methods to add a chained call to a relation method to define it as soft-deleteable.

src/Database/Relations/Concerns/CanBeDetachable.php
bennothommo: Nope, it's been available to Laravel for a while. We also support it through the `detach` property in the current property-style definition.

src/Database/Relations/Concerns/AttachOneOrMany.php
bennothommo: The position wasn't switched - it was simply that the `$relationName` property was removed, and the `$fieldName` property as added afterwards. `$relationName` is now specified in the `Winter\Storm\Database\Relations\Concerns\HasRelationName` concern.

`$fieldName` is required for parity with Laravel's base morph relations, although I don't believe we use it ourselves.

src/Database/Relations/BelongsToMany.php
bennothommo: Subjective, but I feel this style reads better. ie. relation type can be counted, can be detached, etc.

src/Database/Relations/BelongsToMany.php
bennothommo: We do cover it under **Curly braces for condition blocks** here: https://wintercms.com/docs/v1.2/docs/architecture/developer-guide#php-coding-standards

It's probable that I missed this or copied it from some old code.

src/Database/Concerns/HasRelationships.php
bennothommo: It's used to determine the relation name for relations created by calling a method when you've only defined property-style relations. It was in the original version of this file.

Basically, it's so, for example:

```
$comments = $post->comments()->get();
```

When calling `comments()`, if that was defined as a property relation, it uses `comments` as the relation name when creating the relation object, and determining which config to retrieve.

src/Database/Concerns/HasRelationships.php
bennothommo: There's two changes happening here:

1. The original public methods are no longer in our `HasRelationships` definition. These are now provided by Laravel core.
2. The protected `new$relationType` methods are what we extend from the original Laravel definitions, mainly to provide a relation name and some small adjustments for our own needs. I believe these were *not* in the original file, so that's the reason for the messy diff.

src/Database/Relations/Concerns/CanBeExtended.php
LukeTowers: Does this get run only once globally or is it called every time the SoftDelete trait is booted? @bennothommo

src/Database/Relations/BelongsToMany.php
LukeTowers: Do we have other examples in our code (or Laravel's) using the CanBe? Just wondering if it's worth breaking convention.

src/Database/Relations/Concerns/CanBeExtended.php
bennothommo: Good question - I'll need to check this. It might be every time it's booted.

src/Database/Relations/BelongsToMany.php
bennothommo: Laravel uses the prefix for its concerns sometimes, such as [`CanBeOneOfMany`](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Database/Eloquent/Relations/Concerns/CanBeOneOfMany.php) or [`CanBePrecognitive`](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Http/Concerns/CanBePrecognitive.php). I think it's a good way to describe a concern that provides functionality that may be optional.

LukeTowers: Good enough for me!

src/Database/Relations/Concerns/DefinedConstraints.php
LukeTowers:
```suggestion
     * @deprecated 1.3.0 Constraints are now applied on an as-needed basis.
```

LukeTowers (changes requested):

src/Database/Concerns/HasRelationships.php
bennothommo: Fixed.

src/Database/Relations/Concerns/CanBeExtended.php
bennothommo: Added a check to see if the dynamic methods and properties are defined already, and skip their redefinition if so.

LukeTowers (approved):

src/Database/Traits/SoftDelete.php
```diff
@@ -62,6 +75,46 @@ public static function bootSoftDelete()
...
+            HasOne::class,
+            MorphMany::class,
+            MorphOne::class,
+            MorphToMany::class,
+        ] as $relationClass) {
+            $relationClass::extend(function () {
+                // Prevent double-defining the dynamically added properties and methods below
+                if ($this->methodExists('softDeletable')) {
+                    return;
+                }
```
LukeTowers: This still concerns me as I think it becomes an n+1 issue where if a model uses SoftDelete then every instance of that model will cause an additional function to be added to these relation classes, even if that function exits early. Thoughts @bennothommo @jaxwilko?

LukeTowers (changes requested):
Still have performance issues with the softdelete trait

src/Database/Traits/SoftDelete.php
bennothommo: It would be minimised because it would be only on each "class" that uses it, not each model instance. But otherwise, I'm running out of ideas. I tried doing a "local" extension but I think that's too late in the boot process for it to be available in relation methods.

bennothommo:
@LukeTowers I've changed the soft delete methods on the relation to be available at all times, but the actual soft delete flag will only be enabled when `softDeletable()` is called on the relation AND the related model has the `SoftDelete` trait.

src/Database/Dongle.php
```diff
@@ -65,10 +65,6 @@ public function parse($sql)
     public function parseGroupConcat($sql)
     {
         $result = preg_replace_callback('/group_concat\((.+)\)/i', function ($matches) {
-            if (!isset($matches[1])) {
-                return $matches[0];
-            }
```
LukeTowers: Why was this required?

src/Html/Helper.php
```diff
@@ -35,10 +35,6 @@ public static function nameToArray($string)
         }
 
         if (preg_match('/^([^\]]+)(?:\[(.+)\])+$/', $string, $matches)) {
-            if (count($matches) < 2) {
-                return $result;
-            }
```
LukeTowers: Why was this changed?

src/Database/Dongle.php
bennothommo: It's pointless. The first match will always be present if the callback is run, because the callback won't be run if there's no expression match.

src/Html/Helper.php
bennothommo: It'll never be fired. There will always be more than 2 matches.

LukeTowers (approved):

LukeTowers:
@bennothommo is there a related docs PR?

bennothommo:
@LukeTowers yep, here: https://github.com/wintercms/docs/pull/176

bennothommo:
Late change: Using the `Relation` attribute to define a relation method will now require the relation type to be defined as the attribute's first parameter. Only a minor API change, but critical for behind-the-scenes to prevent infinite loops.

So instead of:

```php
#[Relation]
```

It would now be:

```php
#[Relation('hasOne')]
```

src/Database/Concerns/HasRelationships.php
```diff
@@ -14,8 +20,27 @@
 use Winter\Storm\Database\Relations\MorphOne;
 use Winter\Storm\Database\Relations\MorphTo;
 use Winter\Storm\Database\Relations\MorphToMany;
-use Winter\Storm\Support\Str;
-
+use Winter\Storm\Exception\ApplicationException;
```
LukeTowers: SystemException instead please

PR #170, opened 2024-02-29
Fix model.setAttribute event value

### LINKED ISSUE #1047: Model mutators not working using SettingsModel behavior
### Winter CMS Build

dev-develop

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

Hi,

The latest version of WinterCMS is affected by an issue that prevents values returned by mutators to be persisted to the database. I read the `SettingsModel` behavior code and it seems that model `$attributes` are mistakenly overwritten by "original" values.

### Steps to replicate

1. Create a Settings model.
2. Implement a mutator and return a different value for that form field.
3. Read the persisted value.

### Workaround

Apparently the only solution is to override the `beforeSave()` method inside the settings model and to manipulate the JSON contained into `$this->attributes['value']` by first deserializing it and then serializing it back.

Example:
```php
protected function beforeSave()
{
	$fields = json_decode($this->attributes['value'], true);
	$fields['token'] = Crypt::encrypt($fields['token'] ?? '');
	$this->attributes['value'] = json_encode($fields);
}
```

mjauvin:
Fixes wintercms/winter#1047
Replaces https://github.com/wintercms/winter/pull/1071

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1392,7 +1392,7 @@ class Model extends EloquentModel implements ModelInterface
          *     });
          *
          */
-        $this->fireEvent('model.setAttribute', [$key, $value]);
+        $this->fireEvent('model.setAttribute', [$key, $this->getAttribute($key)]);
 
         return $result;
     }
--- a/src/Halcyon/Model.php
+++ b/src/Halcyon/Model.php
@@ -795,7 +795,7 @@ class Model extends Extendable implements ModelInterface, ArrayAccess, Arrayable
          * @see Winter\Storm\Database\Model::setAttributeValue
          */
         // After Event
-        $this->fireEvent('model.setAttribute', [$key, $value]);
+        $this->fireEvent('model.setAttribute', [$key, $this->getAttribute($key)]);
 
         return $this;
     }


mjauvin:
@bennothommo any idea about the larastan error ?

bennothommo:
@mjauvin I have no idea - it seems to be happening randomly at the moment. Retrying the test usually resolves it.

bennothommo (approved):

LukeTowers:
It could just be because it's late and I should probably go to bed, but why exactly is this necessary? This feels very wrong, to be modifying the value returned from the event; why wouldn't it be the job of the listener to retrieve the accessor value if it didn't like the value as provided by this event?

bennothommo:
It goes to the expectation of the event - given that there's a `beforeSetAttribute` event as well, and the value itself is potentially modified from when it first went it to the `setAttribute()` method (it could be trimmed or JSON-encoded or modified by the `beforeSetAttribute` event), the implication is that the `setAttribute` event (being an *after the fact* event) should contain what the attribute is now it has been set, so arguably, this should be after it has been mutated as well.

LukeTowers:
But the setAttribute is dealing with the raw value of the attribute as stored in the attributes array, it's not dealing with the value as potentially mutated through an accessor method; so I'm concerned that this will break things.

Why does it need to be done this way, can't the consumer of this event change what it's doing if it's having problems with that value?

bennothommo:
My main concern is that there's potentially a disconnect between the value that's provided to the event and the value that's actually stored in the attributes array, and I was simply following the behaviour that's being exhibited by these lines:

- https://github.com/wintercms/storm/blob/develop/src/Database/Model.php#L1362
- https://github.com/wintercms/storm/blob/develop/src/Database/Model.php#L1369
- https://github.com/wintercms/storm/blob/develop/src/Database/Model.php#L1376

In that they're modifying the original value set in the parameters (as `$value`) before being sent to the `setAttribute` event.

If the intention of the event was to send the value pre-modification to the `setAttribute` event, then it's not doing that currently and should probably be fixed. If on the other hand, it was to actually be what's stored in the attributes array at that point, then it needs to be modified to include any modification by a mutator, hence changing the event as @mjauvin has proposed.

I'm certainly amenable to adding this as an additional parameter to the event if we must, but I'd rather fix the oversight if possible.

PR #171, opened 2024-03-05
Add support for legacy plugins using October's ExtendableTrait

mjauvin:
Fix infinite loop for Models using October's ExtendableTrait.

### DIFF
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -9,7 +9,6 @@ use Winter\Storm\Support\ClassLoader;
 use Winter\Storm\Support\Serialization;
 use Illuminate\Support\Facades\App;
 use ReflectionException;
-
 use October\Rain\Extension\ExtendableTrait as OctoberExtendableTrait;
 
 /**
@@ -568,7 +567,8 @@ trait ExtendableTrait
             return false;
         }
 
-        while (!in_array(ExtendableTrait::class, $parent->getTraitNames())
+        while (
+            !in_array(ExtendableTrait::class, $parent->getTraitNames())
             && !in_array(OctoberExtendableTrait::class, $parent->getTraitNames())
         ) {
             $parent = $parent->getParentClass();
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -10,6 +10,8 @@ use Winter\Storm\Support\Serialization;
 use Illuminate\Support\Facades\App;
 use ReflectionException;
 
+use October\Rain\Extension\ExtendableTrait as OctoberExtendableTrait;
+
 /**
  * This extension trait is used when access to the underlying base class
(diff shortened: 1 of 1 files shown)


mjauvin:
This was tested with the plugin that was experiencing the infinite loop.

src/Extension/ExtendableTrait.php
```diff
@@ -10,6 +10,8 @@
 use Illuminate\Support\Facades\App;
 use ReflectionException;
 
+use October\Rain\Extension\ExtendableTrait as OctoberExtendableTrait;
```
bennothommo:
```suggestion
use October\Rain\Extension\ExtendableTrait as OctoberExtendableTrait;
```

src/Extension/ExtendableTrait.php
```diff
@@ -566,7 +568,9 @@ protected function extensionGetParentClass(object $instance = null)
             return false;
         }
 
-        while (!in_array(ExtendableTrait::class, $parent->getTraitNames())) {
+        while (!in_array(ExtendableTrait::class, $parent->getTraitNames())
+            && !in_array(OctoberExtendableTrait::class, $parent->getTraitNames())
+        ) {
```
bennothommo:
```suggestion
        while (
            !in_array(ExtendableTrait::class, $parent->getTraitNames())
            && !in_array(OctoberExtendableTrait::class, $parent->getTraitNames())
        ) {
```

src/Extension/ExtendableTrait.php
```diff
@@ -566,7 +567,10 @@ protected function extensionGetParentClass(object $instance = null)
             return false;
         }
 
-        while (!in_array(ExtendableTrait::class, $parent->getTraitNames())) {
+        while (
+            !in_array(ExtendableTrait::class, $parent->getTraitNames())
+            && !in_array(OctoberExtendableTrait::class, $parent->getTraitNames())
```
LukeTowers, line 572: @bennothommo is there any performance concerns about calling getTraitNames() twice instead of storing it in a local variable and checking that variable for both trait names?

bennothommo: I could certainly profile both, but I reckon we'd see very minimal difference between the two. Reflection is very quick in PHP 8 now.

LukeTowers: If you'd be willing to profile both that would be appreciated, given how frequently this could be hit in a request I'd prefer to be better safe than sorry.

bennothommo: The code as it is now (while loop calling `$parent->getTraitNames()` directly):

Screenshot from 2024-03-08 09-36-59

The code updated to store the trait names first and check both traits against the variable:

Screenshot from 2024-03-08 09-41-33

Pretty much exactly the same performance exhibited by both. We're talking nanoseconds of difference.

PR #175, opened 2024-05-03
Override Symfony Process.

mjauvin:
This is required in order to handle an exception when open_basedir restrictions are in effect, preventing a process to be created with TTY mode enabled.

### DIFF
--- a/src/Console/Process.php
+++ b/src/Console/Process.php
@@ -20,7 +20,10 @@ class Process extends BaseProcess
      */
     public function setTty(bool $tty): static
     {
-        if (($openBasedir = ini_get('open_basedir')) && !@is_readable('/dev/tty')) {
+        $isUnix = '/' === \DIRECTORY_SEPARATOR;
+        $openBasedir = ini_get('open_basedir');
+
+        if ($tty && $isUnix && !empty($openBasedir) && !@is_readable('/dev/tty')) {
             throw new RuntimeException("\nYou have PHP open_basedir restrictions enabled.\nTTY mode has been disabled because access to /dev/tty is not allowed.\n");
         }
 
--- a/src/Console/Process.php
+++ b/src/Console/Process.php
@@ -12,7 +12,7 @@ use Symfony\Component\Process\Exception\RuntimeException;
 class Process extends BaseProcess
 {
     /**
-     * Enables or disables the TTY mode.
+     * Enables or disables TTY mode.
      *
      * @throws RuntimeException In case the TTY mode is not supported or /dev/tty is not accessible.
      */
@@ -20,7 +20,7 @@ class Process extends BaseProcess
     {
         if ($tty && '/' === \DIRECTORY_SEPARATOR) {
             try {
-                $readable = is_writable('/dev/tty');
+                $writable = is_writable('/dev/tty');
             } catch (\Throwable $e) {
                 throw new RuntimeException($e->getMessage());
             }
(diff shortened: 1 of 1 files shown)


mjauvin:
We need to test this on windows.

LukeTowers:
Symfony looks like it already does this? https://github.com/symfony/symfony/commit/a1398f6de437c6826b1872f1f89d9ae630ba3c3e

mjauvin:
> Symfony looks like it already does this? [symfony/symfony@a1398f6](https://github.com/symfony/symfony/commit/a1398f6de437c6826b1872f1f89d9ae630ba3c3e)

This code:
```php
    public function setTty(bool $tty): static
    {
        if ('\\' === \DIRECTORY_SEPARATOR && $tty) {
            throw new RuntimeException('TTY mode is not supported on Windows platform.');
        }

        if ($tty && !self::isTtySupported()) {
            throw new RuntimeException('TTY mode requires /dev/tty to be read/writable.');
        }

        $this->tty = $tty;

        return $this;
    }
```
Is not the same at all, if you look carefully.

LukeTowers:
@mjauvin what's the difference? It looks like isTtySupported() also checks the readable / writable status of `/dev/tty`

mjauvin:
> @mjauvin what's the difference? It looks like isTtySupported() also checks the readable / writable status of `/dev/tty`

It does, but it looks like PHP's open_basedire restrictions don't apply for this check, but they do for `is_readable()`

LukeTowers:
Interesting. Can we override isTtySupported() instead and also submit an issue to Symfony to report the problem?

mjauvin:
> Interesting. Can we override isTtySupported() instead and also submit an issue to Symfony to report the problem?

I tried with this overriden method but it doesn't get called somehow:
```php
    public static function isTtySupported(): bool
    {   
        static $isTtySupported;

        if ('/' === \DIRECTORY_SEPARATOR) {
            try {
                $readable = is_readable('/dev/tty');
            } catch (\Exception $e) {
                throw new RuntimeException($e->getMessage());
            }
        }

        return $isTtySupported ??= ('/' === \DIRECTORY_SEPARATOR && stream_isatty(\STDOUT));
    }
```

mjauvin:
Only if I replicate Symfony's `setTty()` method in my overriden class, then my `isTtySupported` method gets called.

Maybe because this is a static method ?

mjauvin:
@LukeTowers The problem is Symfony's setTty() method uses "self::isTtySupported()` instead of `static::isTtySupported()` which prevents my overriden class method to be called.

src/Console/Process.php
```diff
@@ -0,0 +1,33 @@
...
+     * @return $this
+     *
+     * @throws RuntimeException In case the TTY mode is not supported or /dev/tty is not accessible.
+     */
+    public function setTty(bool $tty): static
+    {
+        if ($tty && '/' === \DIRECTORY_SEPARATOR) {
+            try {
+                $readable = is_readable('/dev/tty');
+            } catch (\Exception $e) {
```
bennothommo: Just for completeness (even though it's not strictly required)
```suggestion
            } catch (\Throwable $e) {
```

src/Console/Process.php
```diff
@@ -0,0 +1,33 @@
...
+ * overrides Symfony Process
+ *
+ * @author Marc Jauvin
+ */
+class Process extends BaseProcess
+{
+    /**
+     * Enables or disables the TTY mode.
+     *
+     * @return $this
```
bennothommo: Superfluous
```suggestion
```

bennothommo (approved):

mjauvin:
Symfony merged the PR, so we won't be needing this.

PR #176, opened 2024-05-13
Most relations were missing getRelationName method.

mjauvin:
Add missing getter for protected `relationName` property.


mjauvin:
I wish the different Relation classes would have a common base to add the common properties/methods to... this is all over the place.

mjauvin:
How could we call a new Concern shared by all relations to include the `protected $relationName` / `getRelationName()` to (and other common properties/methods) ?

bennothommo:
@mjauvin I think I did this work already in the Laravel Relations PR. :)

mjauvin:
Included in #167

PR #178, opened 2024-05-14
Fix infinite loop when using HasSortableRelations on a model with a self-referencing relationship

mjauvin:
I have model that has a `belongsToMany` relation on itself and somehow, calling the relation on the model (i.e. `$model->$relationName()`) creates an infinite loop.

Using only the model relation definition solves the issue.

### DIFF
--- a/src/Database/Traits/HasSortableRelations.php
+++ b/src/Database/Traits/HasSortableRelations.php
@@ -57,19 +57,18 @@ trait HasSortableRelations
         });
 
         foreach ($sortableRelations as $relationName => $column) {
-            $relation = $this->$relationName();
-            if (method_exists($relation, 'updateExistingPivot')) {
-                // Make sure all pivot-based defined sortable relations load the sort_order column as pivot data.
-                $definition = $this->getRelationDefinition($relationName);
-                $pivot = array_wrap(array_get($definition, 'pivot', []));
+            $realtionType = $this->getRelationType($relationName);
+            if (!in_array($realtionType, ['belongsToMany', 'morphToMany'])) {
+                continue;
+            }
+            $definition = $this->getRelationDefinition($relationName);
+            $pivot = array_wrap(array_get($definition, 'pivot', []));
 
-                if (!in_array($column, $pivot)) {
-                    $pivot[] = $column;
-                    $definition['pivot'] = $pivot;
-
-                    $relationType = $this->getRelationType($relationName);
-                    $this->$relationType[$relationName] = $definition;
-                }
+            if (!in_array($column, $pivot)) {
+                // Make sure the sort order column is available as pivot data.
+                $pivot[] = $column;
+                $definition['pivot'] = $pivot;
+                $this->$relationType[$relationName] = $definition;
             }
         }
     }
--- a/src/Database/Traits/HasSortableRelations.php
+++ b/src/Database/Traits/HasSortableRelations.php
@@ -57,8 +57,8 @@ trait HasSortableRelations
         });
 
         foreach ($sortableRelations as $relationName => $column) {
-            $realtionType = $this->getRelationType($relationName);
-            if (!in_array($realtionType, ['belongsToMany', 'morphToMany'])) {
+            $relationType = $this->getRelationType($relationName);
+            if (!in_array($relationType, ['belongsToMany', 'morphToMany'])) {
                 continue;
             }
             $definition = $this->getRelationDefinition($relationName);
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -11,6 +11,7 @@ parameters:
         - src/Parse/PHP/ArrayFile.php
         - src/Parse/PHP/ArrayPrinter.php
         - src/Foundation/Console/KeyGenerateCommand.php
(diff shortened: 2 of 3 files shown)


src/Database/Traits/HasSortableRelations.php
```diff
@@ -57,19 +57,18 @@ public function initializeHasSortableRelations() : void
...
-
-                if (!in_array($column, $pivot)) {
-                    $pivot[] = $column;
-                    $definition['pivot'] = $pivot;
-
-                    $relationType = $this->getRelationType($relationName);
-                    $this->$relationType[$relationName] = $definition;
-                }
+            $realtionType = $this->getRelationType($relationName);
+            if (!in_array($realtionType, ['belongsToMany', 'morphToMany'])) {
```
LukeTowers:
```suggestion
            $relationType = $this->getRelationType($relationName);
            if (!in_array($relationType, ['belongsToMany', 'morphToMany'])) {
```

LukeTowers:
> somehow, calling the relation on the model (i.e. $model->$relationName()) creates an infinite loop.

Not a fan of "somehow" issues being solved without understanding exactly how 😜 Can we dig further into why exactly an infinite loop is caused when attempting to access the relation method at this point in time? I would prefer to address that problem (or at least fully understand it) before just giving up and resorting to the relationship definition properties. 

Does this support relations defined via methods instead of the property approach?

mjauvin:
> Not a fan of "somehow" issues being solved without understanding exactly how 😜 Can we dig further into why exactly an infinite loop is caused when attempting to access the relation method at this point in time? I would prefer to address that problem (or at least fully understand it) before just giving up and resorting to the relationship definition properties.

It works for all my models except one that has a relation on itself, so we need to look there.

> Does this support relations defined via methods instead of the property approach?

What do you mean "relations defined via methods" ?

mjauvin:
@LukeTowers I was able to reproduce with a MUCH simpler model and a trait specifically written to only trigger this behavior:
```php
trait Test
{
    public function initializeTest() : void
    {
        $relation = $this->myrelation();
    }   
}
```
And the model:
```php
class MyModel extends \Model
{
    use Test;

    public $table = 'mytable';

    public $belongsToMany = [
        'myrelation' => [
            MyModel::class,
        ],
    ];
}
```

mjauvin:
Not sure EXACTLY what is triggering the loop, but the trait initializes itself then for each relation instanciated (which uses the same model), the trait is probably triggered again when the relation is instanciated...

This can easily be reproduced in `php artisan tinker` by instanciating the model.

```
# new MyModel

PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 262144 bytes) in /var/www/vhosts/iet.studioazura.com/vendor/winter/storm/src/Database/Concerns/HasRelationships.php on line 224
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 262144 bytes) in Unknown on line 0
```

mjauvin:
@LukeTowers I did a bit more digging and it turns out Laravel's HasRelationships concern adds a call to `newRelatedInstance()` method for all relation types which basically create a new instance of the related model... so if the related model is itself (with the HasSortableRelations trait), then that's where the loop gets created.

mjauvin:
So in conclusion, I think the fix in this PR is the right way to address the issue.

And to address your question about relations defined in a method (I assume you meant what @bennothommo introduced in [this PR](#167), the `getRelationDefinition()` method in his PR works for both way to define relations.

LukeTowers:
@mjauvin Sounds good, is this something we can add a unit test for?

mjauvin:
@LukeTowers What exactly do you want to test ?

LukeTowers:
Self relations not causing infinite loops :)

mjauvin:
How do you test for that? `AssertNoInfiniteLoop()` ?

🤣

mjauvin:
I added a unittest and tested outside of this branch and it throws a fatal php error as shown below:
```
$ ./vendor/bin/phpunit tests/Database/Traits/HasSortableRelationsTest.php
PHPUnit 9.6.19 by Sebastian Bergmann and contributors.

PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 262144 bytes) in /var/www/vhosts/wn.r4l.com/winter/storm/vendor/laravel/f
ramework/src/Illuminate/Database/Eloquent/Model.php on line 323
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 262144 bytes) in /var/www/vhosts/wn.r4l.com/winter/storm/vendor/spatie/ra
y/src/helpers.php on line 55
```

When running under this PR, it runs as expected:
```
$ ./vendor/bin/phpunit tests/Database/Traits/HasSortableRelationsTest.php 
PHPUnit 9.6.19 by Sebastian Bergmann and contributors.

.                                                                   1 / 1 (100%)

Time: 00:00.057, Memory: 28.00 MB

OK (1 test, 1 assertion)
```

mjauvin:
@bennothommo @LukeTowers Do we really need to use phpstan? I lose so much time just trying to figureout/fix bogus issues it is not really helping...

PR #180, opened 2024-05-29
Ability to override existing relations 

goldmont:
[Discussion #1139](https://github.com/orgs/wintercms/discussions/1139)

@mjauvin

### DIFF
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -800,7 +800,7 @@ trait HasRelationships
      *
      * @throws InvalidArgumentException if the $type is invalid or if the $name is already in use
      */
-    protected function addRelation(string $type, string $name, array $config): void
+    protected function addRelation(string $type, string $name, array $config, bool $override = false): void
     {
         if (!in_array($type, static::$relationTypes)) {
             throw new InvalidArgumentException(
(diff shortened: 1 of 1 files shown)


LukeTowers:
@goldmont I'm not sure I'm comfortable with this approach. This also bypasses the check for a conflicting property and could have unintended side effects. 

IMO if a developer is going to be overriding an existing relationship then they need to be more explicit with that decision, in which case I think manually interacting with the original relationship definition is the best approach (which is how it was done before this method existed even just to add new relationships).

@bennothommo any thoughts from you on this?

mjauvin:
@LukeTowers I would argue that specifying override:true is explicit enough...

bennothommo:
I'm of the opinion that if you want to use a custom relation class, you add a separate relation. You don't override the original. The issue is that overriding the original basically sets you up for some feature to fail because it was expecting the original relation. You basically are forced to ensure that your custom relation class always sticks to the API of the original class it is overriding.

LukeTowers:
> I would argue that specifying override:true is explicit enough...

I disagree, it encourages developers to always set it. The point of the helper method is to have this error checking in the first place: https://github.com/goldmont/storm/commit/b977610d361cc25f24b05062e40a83a6d53680a8

> Previously the developer would have to set the relation property directly themselves by merging their additions with the existing definitions and usually they wouldn't include any error checking logic

LukeTowers:
Closing for now, will re-open if anyone has anything else to add to the discussion.

PR #184, opened 2024-07-29
Use fromRawAttributes() for custom pivot models

### LINKED ISSUE #741: Repeater field not initialzed properly pivot modal
### Winter CMS Build

dev-develop

### PHP Version

8.0

### Database engine

MySQL/MariaDB

### Plugins installed

Winter.Test

### Issue description

If there is a relation widget with pivot model with a repeater type field, then in the pivot modal the repeater is always empty. If I enter some data to it and save it, the data is properly updated in the database, but upon opening the pivot modal the repeater is stil empty.

### Steps to replicate

Please see this PR: https://github.com/wintercms/wn-test-plugin/pull/11

### Workaround

_No response_

mjauvin:
Fixes https://github.com/wintercms/winter/issues/741

### DIFF
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -853,7 +853,7 @@ class Model extends EloquentModel implements ModelInterface
 
         if (!is_null($definition) && array_key_exists('pivotModel', $definition)) {
             $pivotModel = $definition['pivotModel'];
-            return $pivotModel::fromAttributes($parent, $attributes, $table, $exists);
+            return $pivotModel::fromRawAttributes($parent, $attributes, $table, $exists);
         }
     }
 


mjauvin:
@hirisov if you could test this fix, that would be appreciated.

Thanks.

mjauvin:
@bennothommo this should resolve the issue with the User's Roles awards in Winter.Test plugin as well.

LukeTowers:
@mjauvin does line 838 also need to be changed?

mjauvin:
> @mjauvin does line 838 also need to be changed?

I don't think so, but I'm not 100% sure.

mjauvin:
> @mjauvin does line 838 also need to be changed?

According to the `newPivot()` method in Laravel's Eloquent Model, I think we're good:
```php
    public function newPivot(self $parent, array $attributes, $table, $exists, $using = null)
    {   
        return $using ? $using::fromRawAttributes($parent, $attributes, $table, $exists)
            : Pivot::fromAttributes($parent, $attributes, $table, $exists);
    }       
```

But I realize we could completely remove our overriden method now, since it's a carbon copy of Laravel's ...

mjauvin:
See https://github.com/wintercms/storm/pull/185

LukeTowers:
See https://github.com/wintercms/storm/pull/185#issuecomment-2258638874, our override is necessary to return our version of the Pivot model.

PR #185, opened 2024-07-30
newPivot method is the same as Laravel

mjauvin:
No need to duplicate the same code here.


LukeTowers:
> [use Storm's pivot model](https://github.com/wintercms/storm/pull/185/commits/1f9ecafe01c7aa0eaf5cefcc165a1fa2c951b7b0)

That commit explains why we have this method. The code is the same visually, but by having the override here we are returning a `Winter\Storm\Database\Pivot` which extends `Winter\Storm\Database\Model` instead of a `Illuminate\Database\Eloquent\Relations\Pivot` which extends `Illuminate\Database\Eloquent\Model`.

src/Database/MorphPivot.php
```diff
@@ -148,7 +148,7 @@ public function newQueryForRestoration($ids)
         }
 
         if (!str_contains($ids, ':')) {
-            return parent::newQueryForRestoration($ids);
+            return parent::newQueryForRestoration([$ids]);
```
LukeTowers, line 151: This matches [Laravel's code](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Relations/MorphPivot.php#L130-L152) exactly, so I'm fine not making this change. Really what should be changed is the docblock in Laravel's code.

PR #186, opened 2024-08-01
Restore File::unique() helper

damsfx:
Restore : 
- https://github.com/wintercms/storm/pull/114/commits/68bb7acc6c0a1377b375d3dc9052310a9d8d02d8 (partially)
- https://github.com/wintercms/storm/pull/114/commits/76a4302553cf68a3c955a320a392c8b61bf0e3ba

## Summary by CodeRabbit

* **New Features**
  * Added a new filename uniqueness utility that automatically generates unique names when duplicates are detected. The method appends sequential numbers to conflicting filenames using fully customizable separators (underscore by default) and starting indexes, providing flexible naming conventions. This ensures reliable handling of file naming conflicts.


mjauvin:
@damsfx you also need to add the method signature to the File facade (`winter/storm/src/Support/Facades/File.php`)

damsfx:
> @damsfx you also need to add the method signature to the File facade (`winter/storm/src/Support/Facades/File.php`)

@mjauvin  Just done it!
However, having done this on the `develop` branch was an error ... I'm always afraid in this case of not being in line with the winter repo.

tests/Filesystem/FilesystemTest.php
```diff
@@ -11,6 +11,14 @@ public function setUp(): void
         $this->filesystem = new Filesystem();
     }
 
+    public function testUnique()
+    {
+        $this->assertSame('winter_4.cms', $this->filesystem->unique('winter.cms', ['winter_1.cms', 'test_5', 'winter_3.cms']));
+        $this->assertSame('winter_98.cms', $this->filesystem->unique('winter.cms', ['winter_97.cms', 'test_5', 'winter_1.cms']));
+        $this->assertSame('winter 1.cms', $this->filesystem->unique('winter.cms', ['winter_1.cms', 'test_5', 'winter_3.cms'], ' '));
+        $this->assertSame('winter_1.cms', $this->filesystem->unique('winter.cms', ['test_5']));
```
LukeTowers:
```suggestion
      $cases = [
            // File exists, make it unique
            'winter_1.cms' => ['winter.cms', ['winter.cms', 'test_5']],
            
            // File already unique, return original
            'winter.cms' => ['winter.cms', ['winter_1.cms']],

            // Last index available is incremented
            'winter_4.cms' => ['winter.cms', ['winter_1.cms', 'test_5', 'winter_3.cms']],
            'winter_98.cms' => ['winter.cms', ['winter_97.cms', 'test_5', 'winter_1.cms']],

            // Separator as space
            'winter 1.cms' => ['winter.cms', ['winter_1.cms', 'test_5', 'winter_3.cms'], ' '],
        ];
        foreach ($cases as $output => $config) {
            $this->assertSame($output, $this->filesystem->unique(...$config));
        }
```

src/Filesystem/Filesystem.php
```diff
@@ -231,6 +231,34 @@ public function isLocalDisk(\Illuminate\Filesystem\FilesystemAdapter $disk): boo
         return ($disk->getAdapter() instanceof \League\Flysystem\Local\LocalFilesystemAdapter);
     }
 
+    /**
+     * Apply a unique index to a filename from provided list i.e.
+     * winter.txt, [winter_1.txt, winter_2.txt] -> winter_3.txt
+     * winter.txt, [winter_1.txt, winter_3.txt] -> winter_4.txt
```
LukeTowers, line 237:
```suggestion
     * winter.txt, [winter.txt] -> winter_1.txt
     * winter.txt, [winter.txt, winter_1.txt, winter_2.txt] -> winter_3.txt
     * winter.txt, [winter.txt, winter_1.txt, winter_3.txt] -> winter_4.txt
     * winter.txt, [winter_1.txt, winter_3.txt] -> winter.txt
```

@damsfx it should probably allow the input to be returned unmodified if it doesn't exist within the array.

damsfx: @LukeTowers If the entry doesn't exist in the array, it must be returned with another index anyway, right?

LukeTowers: @damsfx I'm making the suggestion that if it doesn't exist in the array it should be returned unmodified. The method is intended to give you a unique filename provided the input of a desired filename and a list of already existing options. If the filename provided is already unique, then there's no need to modify it.

damsfx: @LukeTowers So in a such case, the same filename is needed in the array of references to get the next incremented name.

LukeTowers: The suggested change above should list all the cases that I think the method should need to handle, I've already updated the tests so it should just require a minor change in the method itself.

damsfx: @LukeTowers Your modifications in tests goes to a failure for me.

```php
// File already unique, return original
'winter.cms' => ['winter.cms', ['winter_1.cms']],
```
**FAILED !**
The function return `wintert_2.cms` in that case.  

What I don't understand is why you want to return the unmodified value if it's unique.
In any case, we should return the incremented value! 🤯
```php
(new Filesystem())->unique('winter.cms', ['test.cms']);   // winter_1.cms
(new Filesystem())->unique('winter.cms', []);             // winter_1.cms
```

LukeTowers: @damsfx The tests are failing because the method doesn't currently have the logic to allow the name to be returned unmodified. The method claims to return a unique filename given the input of a desired filename and a list of existing filenames; why does it have to always return an incremented value if the desired filename isn't present?

damsfx: @LukeTowers `Apply a unique index to a filename from provided list ...`  
Maybe we should then say `Apply a unique index to a filename from provided list ONLY if it is found in the list ...` or change the name of the method which is perhaps the source of the confusion.

In any case, the current function respects the logic of the original one: 
https://github.com/wintercms/storm/commit/68bb7acc6c0a1377b375d3dc9052310a9d8d02d8

damsfx: @LukeTowers Do we continue with this PR in draft?

PR #187, opened 2024-09-01
:sparkles: Adds option to ignore revision records with same old and new values ​​in Revisionable Trait

phpiando:
# Addition of a new revisionIgnoreSameValue property:

This property is a boolean flag that allows you to enable or disable the functionality of ignoring revisions where the old value is equal to the new one. By default, this functionality is disabled (false).

# Implementation of the ignoreRevisionSameValue($oldValue, $newValue) method:

This method checks whether the revisionIgnoreSameValue property is enabled and, if so, returns true when the old value is equal to the new value, thus skipping the creation of the revision record.

# Adjustment to revision creation logic in the revisionableAfterUpdate() method:

Added a check to call the ignoreRevisionSameValue method before adding a revision record.

# Justification for Changes:

- Optimization of Revision Records: By avoiding the creation of unnecessary revisions when values ​​do not change, the storage of revisions becomes more efficient and less polluted.
- Flexibility: The new revisionIgnoreSameValue property allows developers to optionally control this functionality, which can be enabled or disabled according to the needs of the project.

# Additional Notes:
Testing has been performed to ensure that the default behavior of the Revisionable trait has not changed and that revisions are still recorded correctly when this new functionality is disabled.
It is recommended that you review the documentation for additional information about this new option.

# Checklist:

- Code follows the project's style standards.
- Tests have been added or updated to cover the new functionality.
- Documentation has been updated as needed.

### DIFF
--- a/src/Database/Traits/Revisionable.php
+++ b/src/Database/Traits/Revisionable.php
@@ -20,6 +20,12 @@ trait Revisionable
      * public $revisionableLimit = 500;
      */
 
+    /**
+     * @var bool Flag for ignoring revision records with the same old and new values.
+     *
+     * public $revisionIgnoreSameValue = false;
+     */
+
     /*
      * You can change the relation name used to store revisions:
      *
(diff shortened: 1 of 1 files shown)


bennothommo:
Hi @phpiando - thanks for submitting this PR.

My understanding is that this trait already checks if revisionable fields have been changed or not from this line: https://github.com/wintercms/storm/blob/b5f971f9749911ea4ac080c4c4c5123ac87ba2a1/src/Database/Traits/Revisionable.php#L75. The `getDirty()` method should only return fields that have been changed, so your additional check would be redundant at that point.

Do you have an example of the Revisionable trait saving unchanged data in a new revision?

PR #188, opened 2024-09-18
Remove all revisions when revisionable model's record is permanently removed

lex0r:

### DIFF
--- a/src/Database/Traits/Revisionable.php
+++ b/src/Database/Traits/Revisionable.php
@@ -105,7 +105,8 @@ trait Revisionable
             class_uses_recursive(get_class($this))
         );
 
-        if (!$softDeletes) {
+        if (!$softDeletes || $softDeletes && !$this->isSoftDelete()) {
+            $this->revisionableFullCleanUp();
             return;
         }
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
Going to need unit tests for this (covering existing behaviour and the change) before this can be merged.

bennothommo:
Just to note, this is a breaking change. I'd still like it to work similar to [my comments here](https://github.com/wintercms/winter/issues/315#issuecomment-1955662725), especially since this is a potential data loss pain point.

lex0r:
> Just to note, this is a breaking change. I'd still like it to work similar to [my comments here](https://github.com/wintercms/winter/issues/315#issuecomment-1955662725), especially since this is a potential data loss pain point.

Currently a non-soft deletion of a model record leads to data loss regardless of whether revisions are deleted or kept. This patch doesn't aim at enhancing revisions capabilities, it only cleans up the mess left behind the current mechanism.

bennothommo:
@lex0r how so? The only loss of data currently is the model record itself which is the intended action - the revisions are still stored.

I'm sorry if I'm appearing to be terse about this, but anything to do with data deletion, I want strong evidence that it is necessary before we put a change on everyone that will be forced onto them.

As mentioned previously, I'd be a lot more comfortable with this if it were opt-in.

PR #191, opened 2024-10-15
Support project relative paths to sqlite databases regardless of webroot

mjauvin:
This allows setting SQLite db path relative to app root, even when running with mirrored install.


mjauvin:
@LukeTowers should we revert this in the Laravel 11 branch since the fix has been merged upstream?

LukeTowers:
@mjauvin yes please

PR #196, opened 2024-12-03
Add model:prune console command

mjauvin:
To show only active plugins, we need to add an event listener in Winter core as shown below:

```php
Event::listen('system.console.model.prune.findModels', function () {
    $models = [];
    $pm = \System\Classes\PluginManager::instance();
 
    $modulesModels = File::findModels([base_path() . '/modules' => '/*/models']);
    $models[] = $modulesModels->all();
 
    $pluginsPaths = collect($pm->getPlugins())->map(function ($plugin) use ($pm) { 
        return $pm->getPluginPath($plugin);
    })->filter(function ($path) {
        return File::exists($path . '/models');
    })->each(function ($path) use (&$models) {
        $modelPaths = Finder::create()->in($path . '/models')->files()->name('/^[A-Z]{1}.+\.php$/');
        $models[] = collect($modelPaths)->map(function ($model) {
            $modelPath = str_replace(['/', '.php'], ['\\', ''], Str::after($model->getRealPath(), plugins_path().DIRECTORY_SEPARATOR));
            return ucwords($modelPath, '\\');
        })->all();
    });
 
    return collect($models)->flatten();
});

```

### DIFF
--- a/src/Database/Console/PruneCommand.php
+++ b/src/Database/Console/PruneCommand.php
@@ -28,12 +28,7 @@ class PruneCommand extends BasePruneCommand
             throw new InvalidArgumentException('The --models and --except options cannot be combined.');
         }
 
-        $paths = [
-            base_path() . '/modules' => '/*/models',
-            plugins_path() => '/*/*/models',
-        ];
-
-        return File::findModels($paths)
+        return $this->findModels()
             ->when(! empty($except), function ($models) use ($except) {
                 return $models->reject(function ($model) use ($except) {
                     return in_array($model, $except);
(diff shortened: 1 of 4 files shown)


src/Database/Console/PruneCommand.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        try {
+            $uses = class_uses_recursive($model);
+        } catch (Exception $e) {
+            return false;
+        }
+
+        return in_array(Prunable::class, $uses) || in_array(MassPrunable::class, $uses);
+    }
+}
```
LukeTowers, line 80: @mjauvin let's just include this in the System module directly. it's pretty dependent on that layer of Winter rather than being able to standalone in Storm.

mjauvin:
moved to a different PR in winter core

PR #198, opened 2024-12-21
Unset event handler when halting

mjauvin:
This is supposed to run once, make it so in all cases.

### DIFF
--- a/src/Support/Traits/Emitter.php
+++ b/src/Support/Traits/Emitter.php
@@ -162,6 +162,7 @@ trait Emitter
                     continue;
                 }
                 if ($halt) {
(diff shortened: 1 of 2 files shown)


LukeTowers:
@mjauvin is this still needed? Can we have any unit tests added for this?

mjauvin:
It's needed if we want a `bindEventOnce()` call that is also halting to be removed from the event listener array since the method returns before reaching the call that does this when $halting = false.

mjauvin:
I'll see what I can do for a unit test.

mjauvin:
@LukeTowers was able to add a unit test for this.

I tested commenting out the `unset()` statement and it does indeed fail the unit test.

PR #199, opened 2024-12-22
Fix for dynamic methods defined as [class, method]

mjauvin:
The following breaks HasRelationships concern:
```php
$model->addDynamicMethod('myMethod', [static::class, 'myMethod']);
```

### DIFF
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -2,7 +2,6 @@
 
 namespace Winter\Storm\Database\Concerns;
 
-use Closure;
 use InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


bennothommo:
@mjauvin we will need a unit test to determine that a relation defined in the array callback format *can* be detected as a relation - your tests so far only seem to be "failure" states.

mjauvin:
Yes, the tests so far are there to trigger the case that was failing before I made the PR.

mjauvin:
> @mjauvin we will need a unit test to determine that a relation defined in the array callback format _can_ be detected as a relation - your tests so far only seem to be "failure" states.

Done!

bennothommo (approved):

PR #203, opened 2025-02-05
Improve PHP 8.4 support

joaocosta-azores:
PHP 8.4 deprecates implicitly nullable types and applications are recommended to explicitly declare the type as nullable.
Message `Implicitly marking parameter $default as nullable is deprecated, the explicit nullable type must be used instead` is shown when declaring a default null value without declaring null in the type declaration.

To fix this, we use Union Type to declare the type also as nullable:
`string $value = null` becomes `string|null $value = null`

`composer run-script test` has been run to locate the code inside `src/` that needed to be fixed.

PR #204, opened 2025-02-10
Added support for more detailed error logging

jaxwilko:
This PR modifies how the details of an exeception are processed to allow us to display more information on the "Error Log" page.

### DIFF
--- a/src/Foundation/Exception/Handler.php
+++ b/src/Foundation/Exception/Handler.php
@@ -350,7 +350,8 @@ class Handler extends ExceptionHandler
         return array_slice(
             $lines,
             $line - (static::EXECPTION_SNIPPET_LINES / 2),
-            static::EXECPTION_SNIPPET_LINES, true
+            static::EXECPTION_SNIPPET_LINES,
+            true
         );
     }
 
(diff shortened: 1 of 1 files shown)


mjauvin:
Nice!

jaxwilko:
@mjauvin ;)

image
image

LukeTowers (approved):

bennothommo:
@jaxwilko @LukeTowers not sure if this was accounted for in this PR already, but it appears closures have more context now in a backtrace. See https://github.com/wintercms/storm/pull/203#issuecomment-2655562172 for more info - it broke the relation name auto-guessing.

Might be useful though for providing more context in the new backtraces.

jaxwilko:
@bennothommo the issue is unless the `Closure` is passed as an argument, we do not have access to it from the stack trace, for example with this code:

image

We get this trace:

image

If the php logic does pass the `Closure` as an argument, such as:

image

The we would gain access to the object:

image

But we cannot 100% confirm that the `Closure`  executed later in the trace is the `Closure` that was previously passed.

While it would be really cool to add this logic, I'm not sure how we could extract the object reliably enough to consistantly generate the snippet (or even to display any extra data).

PR #205, opened 2025-02-19
Fix PHP8.4 errors with symfony/console

mjauvin:
This resolves the issues we're having with PHP8.4

PHP8.4 support was added in symfony/console 6.4.x

related to https://github.com/wintercms/winter/pull/1306

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -6,6 +6,7 @@ on:
     branches:
       - 1.0
       - 1.1
+      - 1.2
       - develop
 
 concurrency:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,7 @@ jobs:
       max-parallel: 6
       matrix:
         operatingSystem: [ubuntu-latest, windows-latest]
-        phpVersion: ['8.0', '8.1', '8.2', '8.3', '8.4']
+        phpVersion: ['8.1', '8.2', '8.3', '8.4']
       fail-fast: false
     runs-on: ${{ matrix.operatingSystem }}
     name: ${{ matrix.operatingSystem }} / PHP ${{ matrix.phpVersion }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,7 @@ jobs:
       max-parallel: 6
       matrix:
         operatingSystem: [ubuntu-latest, windows-latest]
-        phpVersion: ['8.0', '8.1', '8.2', '8.3']
+        phpVersion: ['8.1', '8.2', '8.3', '8.4']
       fail-fast: false
     runs-on: ${{ matrix.operatingSystem }}
     name: ${{ matrix.operatingSystem }} / PHP ${{ matrix.phpVersion }}
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         }
     ],
     "require": {
-        "php": "^8.0.2",
+        "php": ">=8.1",
         "ext-ctype": "*",
         "ext-curl": "*",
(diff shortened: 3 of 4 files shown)


mjauvin:
I don't think the tests failure have anything to do with this, as we've seen in the past.

@bennothommo any idea ?

LukeTowers:
@mjauvin they must have changed the interface in newer versions of symfony/console. We'll need to make a change to our implementation of that method and make sure that it can support the range of versions specified in composer.json or be more specific with our range of supported versions.

mjauvin:
> @mjauvin they must have changed the interface in newer versions of symfony/console. We'll need to make a change to our implementation of that method and make sure that it can support the range of versions specified in composer.json or be more specific with our range of supported versions.

Not sure what you mean.

LukeTowers:
The error in the tests is:

```
Error: Return type (void) of method Winter\Storm\Console\Command::handleSignal() should be compatible with return type (int|false) of method Symfony\Component\Console\Command\SignalableCommandInterface::handleSignal()
 ------ ------------------------------------------------------------------------------ 
  Line   Console/Traits/HandlesCleanup.php (in context of class                        
         Winter\Storm\Console\Command)                                                 
 ------ ------------------------------------------------------------------------------ 
  42     Return type (void) of method                                                  
         Winter\Storm\Console\Command::handleSignal() should be compatible             
         with return type (int|false) of method                                        
         Symfony\Component\Console\Command\SignalableCommandInterface::handleSignal()  
 ------ ------------------------------------------------------------------------------ 
 ```
 
 This means that handleSignal() method interface has been changed in symfony in the latest versions. So in order to support the latest version of symfony console we need to update our implementation in the HandlesCleanup trait to match the symfony one. Then we need to make sure that the change won't affect our support for earlier versions of symfony console, and if it does, then we need to figure out a more specific version constraint so that we can continue supporting PHP 8.0 to 8.4 with Winter.

LukeTowers:
To summarize:

symfony/console must be at ^6.4 to support PHP 8.4.

symfony/console has incompatible changes to the return type and parameters of `handleSignal` in the `SignalableCommandInterface` interface [introduced in v6.3](https://github.com/symfony/console/commit/138fa49967e8196dc0a91d5f333d07ed76248ae3) (and [further changed in v7.0](https://github.com/symfony/console/commit/8b83b2fc425bd48e3e4e863da3a1c8f74d9dac91#diff-8fcdcf5d8bdf975acb3ae773b840a340f51b5de3e88e5da8c72fab27ba320e4f)). We previously attempted to add support in Storm with https://github.com/wintercms/storm/commit/8fafb79109ade5319a2f72ab1b681c48dbfc102e & https://github.com/wintercms/storm/commit/11fc6c79336d21b4c9e3d8040f79f4f55ff60f86 but had to [revert the changes due to the lack of PHP 8.0 support](https://github.com/wintercms/storm/commit/facb27d7486406306dab491ed3d9eb95245fd0b2) in v6.3 of symfony/console.

In order to merge this PR, we must drop support for PHP 8.0. We can either do that as a point release in 1.2, or hold it for 1.3.

Thoughts @wintercms/maintainers?

mjauvin:
@LukeTowers I think it's more interesting to support PHP 8.4 and drop 8.0, but if we decide to release 1.3 once 1.2.8 is released, it might be better to keep support for 8.0 in 1.2.x since that might create trouble for a few users.

LukeTowers:
@mjauvin it won't cause issues for any users to drop support for 8.0 since composer just won't let them update to the release that drops it, whether that's 1.2.8 or 1.3. I'm not sure how long it's going to take to finish 1.3 though.

mjauvin:
Ok, then let's drop 8.0 starting with 1.2.8

PR #209, opened 2025-03-02
Fix belongsTo relationship defined as required in model backend submission

trepmag:
This PR address the following:

When a model $belongsTo relationship is marked as required in $rules then in the backend submission returns "field is required" event though the relation field is filled.


bennothommo:
@trepmag could you please provide an example of what this intends to fix? If possible, please submit a pull request to the [Test plugin](https://github.com/wintercms/wn-test-plugin) that demonstrates the bug occurring.

PR #211, opened 2025-04-17
Add support for saving pivot data directly via widgets

RamonDiaz31:
I use WinterCMS for various projects. In one of them, I needed to save information  to a model with a belongsToMany relationship, including pivot data.

The Behavior RelationController provided by WinterCMS was not suitable for the specific requirements of this project. 
Upon reviewing the default Relation Widget, I decided to replicate its behavior in a custom widget.

Using getSaveValue(), my widget returns a nested array, where the field name corresponds to the name of the belongsToMany relation. When no extra data is defined in the pivot table, it correctly saves the related IDs to the database.

However, once I added an extra column (columnA) to the pivot table, the system began throwing an error: "Field 'columnA' doesn't have a default value". Setting that column to allow null removed the error, but only the related IDs were stored, columnA was always saved as null.

After several trials and adjustments to the input name structure, I encountered a different error: "Invalid column name 'pivot'". At the time, the input name format I used was:
```
input name="[Model][Relation][relation_id][pivot][columnA]" value="A"
```
Changing it to:
```
input name="[Model][Relation][relation_id][columnA]" value="A"
```
Allowed the value of columnA to be saved successfully to the pivot table. This led me to believe that I could add multiple pivot columns using a similar approach.

Unfortunately, I was wrong. When I tried adding three columns like so:
```
input name="[Model][Relation][relation_id][columnA]" value="A"
input name="[Model][Relation][relation_id][columnB]" value="B"
input name="[Model][Relation][relation_id][columnC]" value="C"
```
WinterCMS threw the following error: "Nested Arrays may not be passed to whereIn method".

Eventually, I found the root cause in the file:
```
/vendor/winter/storm/src/Database/Relations/BelongsToMany.php
```
Specifically, the issue was in the setSimpleValue() function. This function uses whereIn() (which doesn't accept nested arrays) to set the relationship, and uses sync() (which support nested arrays for storing  pivot data ) to save the data.

I made adjustments to this function, enabling widgets to save extra pivot data in belongsToMany relationships. I've successfully tested these changes over the course of a week, and I believe they can be useful to other WinterCMS developers facing similar issues.

For the project where I needed this functionality, I ultimately decided to handle pivot data manually (without a widget) for the time being.

### DIFF
--- a/src/Database/Relations/BelongsToMany.php
+++ b/src/Database/Relations/BelongsToMany.php
@@ -65,14 +65,26 @@ class BelongsToMany extends BelongsToManyBase implements RelationInterface
             return;
         }
 
-        // Convert models to keys
+        // Convert models to keys and check if pivot data exists
         if ($value instanceof Model) {
-            $value = $value->getKey();
+            if ($value->pivot instanceof Pivot) {
+                $value = [$value->getKey() => $value->pivot->toArray()];
+            } else {
+                $value = $value->getKey();
+            }
         } elseif (is_array($value)) {
+            $newValue = [];
             foreach ($value as $_key => $_value) {
                 if ($_value instanceof Model) {
-                    $value[$_key] = $_value->getKey();
+                    if ($_value->pivot instanceof Pivot) {
+                        $newValue[$_value->getKey()] = $_value->pivot->toArray();
+                    } else {
+                        $newValue[$_key] = $_value->getKey();
+                    }
                 }
+            }
+            if (count($newValue) > 0) {
+                $value = $newValue;
             }
         }
 
@@ -81,10 +93,30 @@ class BelongsToMany extends BelongsToManyBase implements RelationInterface
             $value = [$value];
         }
 
+        //Check if value has nested arrays (pivot data)
+        $keys = $value;
+        $hasPivot = false;
+        foreach ($value as $_key => $_value) {
+            if (is_array($_value)) {
+                $keys[$_key] = $_key;
+                $hasPivot = true;
+            }
+        }
+
         // Setting the relationship
         $relationCollection = $value instanceof Collection
             ? $value
-            : $relationModel->whereIn($relationModel->getKeyName(), $value)->get();
+            : $relationModel->whereIn($relationModel->getKeyName(), $keys)->get();
+
+
+        // Associate in memory immediately with pivot data (pivot of relation needs to use Winter\Storm\Database\Pivot)
+        if ($hasPivot) {
+            foreach ($relationCollection as  $_relationModel) {
+                if (isset($value[$_relationModel->id])) {
+                    $_relationModel->setRelation('pivot', $this->newPivot($value[$_relationModel->id]));
+                }
+            }
+        }
 
         // Associate in memory immediately
         $this->parent->setRelation($this->relationName, $relationCollection);
--- a/src/Database/Relations/BelongsToMany.php
+++ b/src/Database/Relations/BelongsToMany.php
@@ -65,10 +65,10 @@ class BelongsToMany extends BelongsToManyBase implements RelationInterface
             return;
         }
 
-        // Convert models to keys and check if pivot data exists
+        // if pivot data doesn't exists, Convert models to keys , else convert models to nested arrays with the keys as the indexes (pivot of relation needs to use Winter\Storm\Database\Pivot)
         if ($value instanceof Model) {
-            if ($value->pivot instanceof Pivot) {
-                $value = [$value->getKey() => $value->pivot->toArray()];
+            if ($value->{$this->getPivotAccessor()} instanceof Pivot) {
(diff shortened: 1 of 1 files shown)


damsfx:
I also had issues recording pivot data from the Winter.Test plugin.  
https://github.com/wintercms/wn-test-plugin/issues/15

RamonDiaz31:
The Test plugin utilizes RelationController Behavior, It only works on specific ways, one of the main reasons I discarded it was because I needed something similar to a POS, but not a POS, you put a code in a input and it searches on the database and if exists the pivot data auto-fills, and also because it cannot add a relation with pivot data on create, only on update, when the parent_id is available, else it throws a error:
```
Integrity constraint violation: 1048 Column 'parent_id' cannot be null
```
I also analyzed the file 
```
/modules/backend/behaviors/RelationController.php
```
When I was searching for a solution to my problem.

I indeed find out that:
```
public function onRelationManagePivotCreate()
    {
        $this->beforeAjax();

        /*
         * If the pivot model fails for some reason, abort the sync
         */
        Db::transaction(function () {
            /*
             * Add the checked IDs to the pivot table
             */
            $foreignIds = (array) $this->foreignId;
            $saveData = $this->pivotWidget->getSaveData();
            $foreignModels = $this->relationModel->whereIn($this->relationModel->getKeyName(), $foreignIds)->get();
            $this->relationObject->syncWithPivotValues($foreignModels, $saveData['pivot'] ?? [], false);

            /*
             * Save data to models
             */
            $foreignKeyName = $this->relationModel->getQualifiedKeyName();
            $hydratedModels = $this->relationObject->whereIn($foreignKeyName, $foreignIds)->get();

            foreach ($hydratedModels as $hydratedModel) {
                $modelsToSave = $this->prepareModelsToSave($hydratedModel, $saveData);
                foreach ($modelsToSave as $modelToSave) {
                    $modelToSave->save(null, $this->pivotWidget->getSessionKey());
                }
            }
        });

        return ['#'.$this->relationGetId('view') => $this->relationRenderView()];
    }
```
Never uses DeferredBinding like the other functions of the other types of relationships do.
On my tests I also changed that function to use DeferredBinding and it worked on create, but other functions related to the relationships BelongsToMany/MorphToMany need to be changed and tested, because I think it’s supposed to work only on update.
I searched for information on the subject and found this: https://github.com/wintercms/winter/issues/458
but it just doesn’t work well even when making the id  nullable, it only saved with the correct id on create when I changed the function to use DeferredBinding, even if the parent_id was not nullable.
When I have more time available I’m going to do more tests on that.

My PR allows to insert pivot data with a Widget, mostly custom ones, because I think the Relation Widget is the closest and easier to extend from the official ones that already can save data in a BelongsToMany Relation without the  RelationController Behavior, the only requirements is that the pivot model needs Winter\Storm\Database\Pivot and the relation needs to declare the pivotModel, it doesn't work with the generic one and of course the inputs fields must be in a specific format: 

Example:
```
input name="Order[products][product_id][price]"
input name="Order[products][product_id][qty]"
input name="Order[products][product_id][discount]"
```

So when gathered, the format of the array is:
```
         [
            1 => [
                'price' => 100,
                'qty' => 1,
                'discount' => '50'
            ],
            2 => [
                'price' => 200,
                'qty' => 2,
                'discount' => '20'
            ],
            3 => [
                'price' => 300,
                'qty' => 3,
                'discount' => '10'
            ]
        ];

```
That format is compatible with the sync() method to save the relation ids with additional data.

LukeTowers:
@RamonDiaz31 I'm not sure I fully understand the usecase, are you able to provide a screenshot of how your custom widget functions and any additional details on why the relationcontroller doesn't work for you? 

At the very least this needs test coverage added and possibly a usage example added to the docs or the test plugin.

RamonDiaz31:
Well, about the RelationController, it works very well when not using a belongsToMany Relation with pivot data. I’m posting my primary problem with it.
RelationController01

I’m using the Winter.Test plugin, even here the relationController with Pivot doesn’t work when the parent model is yet to be created. 
RelationController02

Setting the user_id to nullable and remove the primary key,  just make the pivot table save date with null user_id, and never add the correct id even after save the parent. Roles never fills and sends validation error.
RelationController03
RelationController04
RelationController05

Removing the validation of Roles, it can save, but the user_id stays  null.
It works when using DeferredBinding, but it requires to modify the RelationController and I think that’s another topic of discussion. 
RelationController06
RelationController07

I also wanted to share this quick plugin with a custom widget to show how works my PR.
The widget is based on the Relation Widget, which can save data to a BelongsToMany Relationship and even a single column of pivot data, the functionality was already there, I just changed a few lines to make it work with more than one columns of pivot data. https://github.com/RamonDiaz31/SimpleCart
Without my changes it throws the error of whereIn.

03 - Default

With my changes it can create and save successfully. 
04 - With modifications
05 - With Modifications edit

Of course, I'm open to make docs and show more tests if necessary.

PR #212, opened 2025-04-29
Update ->original  to ->originalIsEquivalent

AIC-BV:
Other correct notation is $model->getOriginal('title') but $model->originalIsEquivalent('title) does the compare check by default.
->original accesses a protected property which, as I've been told, is bad practice


mjauvin:
Not only is it bad practice, but in this case, it just wouldn't work since it's not used within a method of the object so the property is not even accessible.

LukeTowers (approved):

PR #214, opened 2025-05-09
Using hasXThrough in conjunction with soft delete

zimudec:
Currently, hasManyThrough allows access to a record from a source table to a target table: table1 -> hasmany -> table_middle -> hasmany -> table2.

It now correctly detects when there are items deleted by soft delete in table_middle.

Due to the wintercms query cache, I had to use "Db::flushDuplicateCache()" during testing to retrieve table1 with the results from table2. I don't know how to avoid this problem.

- Using hasManyThrough in conjunction with soft delete on the intermediate table now respects trashed items
- Testing for hasManyThrough in conjunction with soft delete on the intermediate table

### DIFF
--- a/src/Database/Relations/HasManyThrough.php
+++ b/src/Database/Relations/HasManyThrough.php
@@ -33,7 +33,7 @@ class HasManyThrough extends HasManyThroughBase
      *
      * @return bool
      */
-    public function parentSoftDeletes()
+    public function throughParentSoftDeletes()
     {
         $uses = class_uses_recursive(get_class($this->parent));
 
--- a/src/Database/Relations/HasOneThrough.php
+++ b/src/Database/Relations/HasOneThrough.php
@@ -32,7 +32,7 @@ class HasOneThrough extends HasOneThroughBase
      *
      * @return bool
      */
-    public function parentSoftDeletes()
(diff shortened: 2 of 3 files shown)


LukeTowers:
@zimudec does this change also need to be applied to hasOneThrough?

bennothommo:
@LukeTowers `HasOneThrough` extends `HasManyThrough` :)

bennothommo (approved):

LukeTowers:
@bennothommo but the original method still exists in HasOneThrough: https://github.com/wintercms/storm/blob/develop/src/Database/Relations/HasOneThrough.php#L35-L41. Should it be removed as a part of this PR?

bennothommo:
@LukeTowers ah yeah you're right, must've been looking at Laravel's base relation without realising.

zimudec:
@LukeTowers @bennothommo  I added a change and tests for HasOneThrough as well.

HasOneThrough extends HasManyThrough, but from Eloquent, not from Wintercms. Therefore, I applied the same change to HasOneThrough instead of deleting it.

bennothommo (approved):

LukeTowers (approved):

PR #215, opened 2025-05-16
Check if concats exists as a property on the querybuilder

Pasquale-CS:
This is a workaround but not a solution. Please accept the solution until you find a solution.


Pasquale-CS:
[BUG] issues #1355 open on "wintercms/winter"

LukeTowers:
Please respond to https://github.com/wintercms/winter/issues/1355#issuecomment-2887748073

PR #216, opened 2025-05-17
Use pivot model table name if not defined in the relation

mjauvin:
If using a custom pivot model, the table name may be omitted in the relation definition if it is set in the custom pivot model itself.

Laravel will resolve the table name to the Pivot model table.
(ref. https://github.com/laravel/framework/blob/12.x/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php#L184-L201)

### DIFF
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -440,7 +440,7 @@ trait HasRelationships
             case 'belongsToMany':
                 $relation = $this->belongsToMany(
                     $relatedClass,
-                    $definition['pivotModel'] ?? $definition['table'] ?? null,
+                    $definition['table'] ?? $definition['pivotModel'] ?? null,
                     $definition['key'] ?? null,
                     $definition['otherKey'] ?? null,
                     $definition['parentKey'] ?? null,
--- a/src/Database/Concerns/HasRelationships.php
+++ b/src/Database/Concerns/HasRelationships.php
@@ -481,7 +481,7 @@ trait HasRelationships
                 $relation = $this->morphToMany(
                     $relatedClass,
                     $definition['name'] ?? $relationName,
-                    $definition['pivotModel'] ?? $definition['table'] ?? null,
+                    $definition['table'] ?? $definition['pivotModel'] ?? null,
                     $definition['key'] ?? null,
                     $definition['otherKey'] ?? null,
                     $definition['parentKey'] ?? null,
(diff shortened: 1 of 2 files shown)


src/Database/Concerns/HasRelationships.php
```diff
@@ -440,7 +440,7 @@ protected function handleRelation(string $relationName, bool $addConstraints = t
             case 'belongsToMany':
                 $relation = $this->belongsToMany(
                     $relatedClass,
-                    $definition['table'] ?? null,
+                    $definition['table'] ?? $definition['pivotModel'] ?? null,
```
bennothommo, line 443: Arguably, a pivot model definition should be considered more authoritative than a table definition, on the off-chance that someone specifies both in the config.
```suggestion
                    $definition['pivotModel'] ?? $definition['table'] ?? null,
```

mjauvin: There's nothing that ensures at this point that the pivot model defines the table...

bennothommo: @mjauvin not sure what you mean? If you specify a pivot model, Laravel will determine the table of that model in the standard way (either the `$table` property of the model, or conventionally using the model's name).

mjauvin: Yes, but there's no way to know at that point if the pivot actually defines a table name.

If it doesn't, laravel's way of guessing the table name is not compatible with ours

bennothommo: What's an example of this not working?

We already have a [call to the `using` method](https://github.com/wintercms/storm/blob/develop/src/Database/Concerns/HasRelationships.php#L450-L452) underneath where you've made your edit, which should switch the relation to use the pivot's table.

mjauvin: @bennothommo I changed the order as you suggested, and added a [failing] test to demonstrate my point.

bennothommo: @mjauvin sorry, but I'm still not understanding how this is an issue?

Your failing test is creating a BTM relation called `pivot_without_table` and using the `CustomPivotWithoutTable` class as a custom pivot without defining a table, and then expecting the table name to be `custom_pivot_table`, when by the normal automatic table name logic, it should be `custom_pivot_without_table` which is correctly being returned by the `getTable()` method.

~~It should be said that when defining a BTM relation in Laravel, you define either the table OR the pivot, you do not define both. If your pivot needs to use a table name that doesn't match the convention, then you specify the table name in the pivot.~~

EDIT: Never mind, it was buried deep in Laravel but evidently I'm wrong. The relation can still set the table name for a custom pivot.

bennothommo: I've updated the test to show that it passes with the expected table name.

bennothommo (approved):

PR #220, opened 2025-09-19
Limiting scope of signal handling based on environment

bennothommo:
- Do not handle Linux signals if the PCNTL extension is unavailable. Artisan commands can still function without this extension - they just simply cannot handle termination signals. This prevents commands from throwing an error due to the missing SIGINT/SIGTERM/SIGQUIT constants.
- Prevent multiple registrations of the Windows signal handler if the  method is called multiple times.
- Remove specific checks for Windows signals - the handler only accepts two types of signals (CTRL+C or CTRL+BREAK) and both are termination signals.

## Release Notes

* **Bug Fixes**
  * Improved signal handler registration and cleanup mechanisms for more reliable application termination behavior across Windows and Unix-like systems.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/Console/Traits/HandlesCleanup.php
+++ b/src/Console/Traits/HandlesCleanup.php
@@ -1,4 +1,6 @@
-<?php namespace Winter\Storm\Console\Traits;
+<?php
+
+namespace Winter\Storm\Console\Traits;
 
 /**
  * Console Command Trait that injects cross-platform signal handling to trigger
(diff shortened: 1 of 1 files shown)


LukeTowers:
@bennothommo this looks fine to me, any issues with it going in 1.2.9? Also what's the PHPStan issue?

bennothommo:
@LukeTowers no reason not to include it in 1.2.9 if you would prefer. The PHPStan issue seems to have occurred with all pushes recently to Storm - it doesn't seem to be able to correctly work out how our Singletons work. We can probably ignore it for now.

PR #228, opened 2026-06-01
Support Laravel 13

austinderrick:
Builds on #207 (Laravel 12 support), continuing the 1.3 development line. This PR adds **Laravel 13** support on top of the Laravel 12 work.

**Required by wintercms/winter#1487** (WinterCMS Laravel 13 support).

## Summary

Targets Laravel 13 (released 2026-03-17). Storm re-implements/overrides a number of framework internals, so this covers the framework-level deltas in addition to the dependency bumps.

### Dependency changes (`composer.json`)
- `php` `^8.2` → `^8.3` (Laravel 13 minimum)
- `laravel/framework` `^12.30.1` → `^13.0`
- `laravel/tinker` `^2.8.1` → `^3.0`
- `nesbot/carbon` `^3.0` → `^3.8.4`
- `symfony/console`, `symfony/process`, `symfony/yaml` → `^7.4|^8.0`
- `orchestra/testbench` `^10.0` → `^11.0`
- `phpunit/phpunit` `^11.0` → `^11.5.50` (Laravel 13 + testbench 11 accept PHPUnit 11.5.50; kept on 11.x because `meyfa/phpunit-assert-gd` and `dms/phpunit-arraysubset-asserts` do not yet support PHPUnit 12)

### Code changes
- **`ArraySource` trait** — Laravel 13 throws a `LogicException` when a model is instantiated while it is still booting. `bootArraySource()` previously did `new static` (and `static::insert()`, which also instantiates) during boot. The datasource setup is now deferred and performed lazily on the first `resolveConnection()` call, by which point the model has finished booting.
- **Relation classes** (`HasOne`, `HasMany`, `HasOneThrough`, `HasManyThrough`, `BelongsTo`, `MorphTo`, `MorphOne`, `BelongsToMany`, `MorphToMany`) — added `@template`/`@extends` generics mirroring Laravel 13's relation generics so the `Builder<TRelatedModel>` passed by `HasRelationships` matches the relation constructors (fixes PHPStan covariance errors).
- **`Foundation/Console/DownCommand`** — `option('status', 503)` passed an unsupported 2nd argument to `Command::option()`; the default is already defined by the parent `down` command, so the stray argument was removed.
- **`Auth/Models/Preferences::findRecord()`** — corrected the scope typing so it returns `self|null` rather than `stdClass|null`.
[...]

PR #229, opened 2026-06-02
Fix PHP 8.4 and 8.5 deprecations

austinderrick:
## Summary

Fixes the PHP **8.4** and **8.5** deprecations.

### Fixes
- **PHP 8.5 — null as array offset** (`Using null as an array offset is deprecated, use an empty string instead`):
  - `src/Config/Repository.php` — `afterLoad[$namespace]` (read/isset/write) now uses `$namespace ?? ''` (a null/global namespace is a valid case). *(This is the one reported.)*
  - `src/Halcyon/Processors/Processor.php` — `$items[$fileName]` / `[$fileName => …]` now uses `$fileName ?? ''` (`array_get($result, 'fileName')` may return null).
- **PHP 8.4 — implicit-nullable parameter**:
  - `src/Console/Traits/ProcessesQuery.php` — `int $limit = null` → `?int $limit = null`.
- **PHP 8.1+ (also deprecated on 8.5), removed in PHP 9 — `strftime()`**:
  - `src/Html/FormBuilder.php::selectMonth()` — replaced `strftime()` with `date()`; the `$format` default changes from `'%B'` to `'F'` (date() syntax). Default output (full month names) is unchanged; callers passing custom `strftime`-style formats must switch to `date()` syntax.
- **Test fix (for green CI):** `tests/Scheduling/ScheduleListCommandTest` — updated the invokable-class `schedule:list` expectation (modern Laravel renders the bare class name without the `Closure at:` prefix). This was a pre-existing failure on `wip/1.3` unrelated to the deprecations.

## Testing
- `vendor/bin/phpunit` on **PHP 8.5.5**: 711 tests, 0 failures, 0 Storm-source PHP deprecations.
- `vendor/bin/phpunit` on **PHP 8.4.14**: 711 tests, 0 failures, 0 Storm-source PHP deprecations.
- `phpcs`: clean on the changed files.

## Out of scope
- `vendor/assetic/framework` triggers a `$http_response_header` deprecation on 8.5 — that's a third-party dependency, not Storm.
- PHPStan currently reports pre-existing errors on `wip/1.3` (a larastan-version drift, in the relation generics / baseline) that are unrelated to this PR; they are addressed for the Laravel 13 line in #228.


LukeTowers:
@austinderrick can you submit fixes to assetic/framework as well? We maintain that library.

src/Config/Repository.php
```diff
@@ -194,7 +194,7 @@ protected function load($group, $namespace, $collection)
         // If we've already loaded this collection, we will just bail out since we do
         // not want to load it again. Once items are loaded a first time they will
         // stay kept in memory within this class and not loaded from disk again.
-        if (isset($this->afterLoad[$namespace])) {
+        if (isset($this->afterLoad[$namespace ?? ''])) {
```
LukeTowers: How would namespace be null? Docblock currently only says string, what calls this method and passes it a null? Can we update this method to use real argument types?

src/Config/Repository.php
```diff
@@ -314,7 +314,7 @@ public function package($namespace, $hint)
      */
     public function afterLoading($namespace, Closure $callback)
     {
-        $this->afterLoad[$namespace] = $callback;
+        $this->afterLoad[$namespace ?? ''] = $callback;
```
LukeTowers: Same question

src/Config/Repository.php
```diff
@@ -211,7 +211,7 @@ protected function load($group, $namespace, $collection)
      */
     protected function callAfterLoad($namespace, $group, $items)
     {
-        $callback = $this->afterLoad[$namespace];
+        $callback = $this->afterLoad[$namespace ?? ''];
```
LukeTowers: Same question

src/Halcyon/Processors/Processor.php
```diff
@@ -19,7 +19,7 @@ public function processSelectOne(Builder $query, $result)
 
         $fileName = array_get($result, 'fileName');
 
-        return [$fileName => $this->parseTemplateContent($query, $result, $fileName)];
+        return [($fileName ?? '') => $this->parseTemplateContent($query, $result, $fileName)];
```
LukeTowers: Under what condition would this be null? That seems like a bigger issue and one that should be protected against by returning null earlier from this method.

src/Halcyon/Processors/Processor.php
```diff
@@ -39,7 +39,7 @@ public function processSelect(Builder $query, $results)
 
         foreach ($results as $result) {
             $fileName = array_get($result, 'fileName');
-            $items[$fileName] = $this->parseTemplateContent($query, $result, $fileName);
+            $items[$fileName ?? ''] = $this->parseTemplateContent($query, $result, $fileName);
```
LukeTowers: Under what condition would this be null? Seems like it would be a bigger issue if it did and would also cause issues with overwriting items if multiple returned null filename

src/Html/FormBuilder.php
```diff
@@ -478,12 +478,12 @@ public function selectYear(string $name, int $begin = 1900, ?int $end = null, st
     /**
      * Create a select month field.
      */
-    public function selectMonth(string $name, string|array|null $selected = null, array $options = [], $format = '%B'): string
+    public function selectMonth(string $name, string|array|null $selected = null, array $options = [], $format = 'F'): string
```
LukeTowers, line 481: The Form facade would also need to be updated for this.

LukeTowers (dismissed):

src/Config/Repository.php
austinderrick: The namespace is null for any key without a :: segment. KeyParser::keyParserParseBasicSegments() returns [null, $group, $item] for those, so config('app.debug') arrives here with a null namespace. The old @param string was just wrong; it should have been ?string.

I swapped the ?? '' for a real ?string type hint and a null guard. After-load callbacks only ever get registered against a real namespace, through afterLoading() and package(), so the null global namespace can never have one. Guarding on $namespace !== null before the isset states that invariant directly and sidesteps the null-offset deprecation. callAfterLoad() only runs past that guard, so I typed its parameter string.

src/Config/Repository.php
austinderrick: This is the same as the thread on load(): a null namespace just means a non-namespaced key, which is the common case. I fixed it with a ?string type and a null guard instead of the ?? ''.

src/Halcyon/Processors/Processor.php
austinderrick: That is a fair point. A result with no fileName is bad data, not something to coerce away. It now returns null early instead, which lines up with the method's array|null return type and the null check already at the top. Halcyon datasources always set fileName in practice, so this just makes the invariant explicit rather than a case we expect to hit.

src/Html/FormBuilder.php
austinderrick: Done! I updated the selectMonth annotation on the Form facade to the new 'F' default and string|array|null $selected. The selectRange annotation just above it had also drifted from its real signature, so I fixed that one in the same commit.

src/Halcyon/Processors/Processor.php
austinderrick: Updated!

src/Config/Repository.php
austinderrick: Updated! See above.

austinderrick:
> @austinderrick can you submit fixes to assetic/framework as well? We maintain that library.

Oh, perfect! Submitted! Just a few over there: https://github.com/assetic-php/assetic/pull/47

LukeTowers (approved):

PR #231, opened 2026-06-11
Performance improvements: save flows for NestedTree

austinderrick:
## Summary

Two changes to `NestedTree` that remove most of the query overhead from save and move operations, with no change to the resulting tree state.

**1. `moveTo()` no longer saves every descendant individually.**

A move previously ended with:

```php
foreach ($this->newQuery()->allChildren()->get() as $descendant) {
    $descendant->save();
}
```

Each of those saves exists only to realign `nest_depth`, but pays for a full model lifecycle: `setDepth()` opens a transaction, runs a `reload()`, an ancestor `count()`, and an `UPDATE`, and the save fires every bound model event. Since a subtree moves intact, every descendant shifts by the same depth difference as the moved node, so a single bulk `UPDATE nest_depth = nest_depth + (delta)` over the subtree range replaces the loop. If the moved node had no stored depth (corrupt or legacy data), the code falls back to realigning each descendant from its own ancestry via `setDepth(true)`.

**2. `setDepth()` skips recomputing when the save did not move the node.**

`setDepth()` runs on every `model.afterSave`. For an ordinary save (e.g. renaming a node) it still paid the transaction + reload + ancestor count + UPDATE. It now returns early when the save cycle did not move the node (`moveToNewParentId === false`) and a depth is already stored. `moveToNewParent()` also marks the pending move as processed after handling it, which removes a second, redundant depth recompute that previously followed every move inside the same `afterSave`.

## Measurements

Against a real application table (MySQL/MariaDB), moving a node with 15 descendants via `makeChildOf()`:

| | SQL statements | wall time |
|---|---|---|
| before | 188 | 190ms |
| after | 7 | 6ms |

The statement count is now constant regardless of subtree size (previously ~12 statements *per descendant*, plus whatever the host application binds to its save events; our application clears caches in `afterSave`, so a single drag in a reorder UI also triggered one cache clear per descendant).

A plain `save()` with no parent change drops the trait's 3-statement depth recompute (and its transaction) entirely.

## Behavior notes

- Descendant models no longer fire save events when an ancestor is moved. Those saves were dirty-empty (no attribute changes besides the depth realignment), but applications listening for them would no longer be notified. Flagging for the 1.3 release notes.
- A save that does not change a node's parent no longer re-verifies `nest_depth` against the ancestry, so a manually corrupted depth value is no longer silently self-healed by unrelated saves. Moves still realign depth from the ancestry.

## Tests

Four new tests in `NestedTreeTest`: subtree moves realign descendant depths (deeper and shallower, via `makeChildOf()`/`makeRoot()` and via `parent_id` reassignment + save), non-move saves skip the depth recompute, and a query-count regression test pinning the constant-statement move. Full suite passes: 715 tests, 3145 assertions (13 pre-existing PHPUnit deprecations, 9 pre-existing skips).

### DIFF
--- a/src/Database/Traits/NestedTree.php
+++ b/src/Database/Traits/NestedTree.php
@@ -167,6 +167,10 @@ trait NestedTree
         elseif ($parentId !== false) {
             $this->makeChildOf($parentId);
         }
+
+        // The pending move (if any) has been processed and its depth realigned by moveTo(),
+        // so the depth pass that follows in the model.afterSave event can be skipped.
+        $this->moveToNewParentId = false;
     }
 
     /**
@@ -781,10 +785,19 @@ trait NestedTree
     /**
      * Sets the depth attribute
      *
+     * @param bool $force Recompute the depth even when this save did not move the node.
      * @return \Winter\Storm\Database\Model
      */
-    public function setDepth()
+    public function setDepth($force = false)
     {
+        /*
+         * Recomputing the depth costs a reload and an ancestor count inside a transaction on every
+         * save. Skip it when this save cycle did not move the node and a depth is already stored.
+         */
+        if (!$force && $this->moveToNewParentId === false && $this->getDepth() !== null) {
+            return $this;
+        }
+
         $this->getConnection()->transaction(function () {
             $this->reload();
 
@@ -979,6 +992,12 @@ trait NestedTree
         }
 
         /*
+         * Capture the depth before the move. The whole subtree moves intact, so every descendant
+         * shifts by the same depth difference as this node.
+         */
+        $oldDepth = $this->getDepth();
+
+        /*
          * Perform move
          */
         $this->getConnection()->transaction(function () use ($target, $position) {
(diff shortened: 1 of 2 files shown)


austinderrick:
I tested this against the real NestedTree trait in a stock Winter app on MariaDB. Moving a subtree of 200 nodes runs 608 SQL statements and takes 116 ms on develop. The old code saves every descendant one by one. Each save opens a transaction, reloads, counts ancestors, and writes an update. The bulk depth update in this PR brings the same move down to 8 statements and 2.9 ms, about 37 times faster.

The statement count also stays flat as the subtree grows. At 15, 50, and 200 descendants it is always 8, where develop ran 53, 158, and 608. A plain rename drops from 4 statements to 1, because the depth recompute now skips a node that did not move. There is one honest caveat about the numbers. The 188 to 7 figure in the description came from an app that clears caches in afterSave for each descendant. That alone roughly tripled the count. Stock storm sits at 53 to 8. The linear to constant shape holds either way.

Screenshot 2026-06-24 at 4 43 27 PM

[probe_nested.php](https://github.com/user-attachments/files/29311899/probe_nested.php)
[results.csv](https://github.com/user-attachments/files/29311900/results.csv)
[index.html](https://github.com/user-attachments/files/29311903/index.html)

PR #233, opened 2026-06-11
Bind attach relation keys as strings to use the attachment_id index

austinderrick:
## Problem

The `files` table stores `attachment_id` as an indexed **string** column, but `AttachOneOrMany::addConstraints()` and `addEagerConstraints()` bind the parent key(s) using their native integer type. Comparing a string column against an integer forces the database to coerce the column value for every candidate row, which makes the `attachment_id` index unusable.

On MySQL/MariaDB this is easy to demonstrate with `EXPLAIN`: with an integer binding the planner can only use the low-cardinality `attachment_type` index, so every attachment query scans **every file row of that morph type**. With a string binding it uses both indexes (`ref|filter` with rowid intersection) and examines only the matching rows. On installations with large `system_files` tables this is the difference between a handful of rows examined and tens of thousands — per attachment query, multiplied by however many `attachOne`/`attachMany` relations a page touches.

The eager path is affected the same way: for integer parent keys Laravel selects `whereIntegerInRaw()`, which inlines raw integers into the SQL.

## Fix

- `addConstraints()` casts the parent key to a string (preserving `null`, so the `= null` → `whereNull` conversion is unchanged).
- `addEagerConstraints()` builds the `whereIn` with string-cast keys instead of deferring to the parent implementation. Null keys are preserved as-is so they match nothing, rather than being cast to `''` (which could match orphaned rows). The morph type and field constraints are applied exactly as before.

`getRelationExistenceQuery()` already handles this same mismatch for `has()` queries by casting the parent column to TEXT via `DbDongle::cast`, so this brings the regular constraint paths in line with the approach already used there.

Result matching is unaffected: PHP normalizes numeric-string array keys, so the eager-load dictionary lookup behaves identically (covered by a new end-to-end test).

## Testing

Four new tests in `AttachOneTest`: string bindings on lazy constraints, string bindings on eager constraints, null-key preservation in eager constraints, and an end-to-end eager load asserting attachments still match their parents. Full test suite passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed attachment relationship constraint binding to use string-typed parent keys, ensuring proper index usage and column type matching.
  * Corrected eager-loading behavior for attachment relationships to preserve null parent keys and properly filter results.

* **Tests**
  * Added comprehensive test coverage for attachment relationship constraints and eager-loading scenarios.

### DIFF
--- a/src/Database/Relations/Concerns/AttachOneOrMany.php
+++ b/src/Database/Relations/Concerns/AttachOneOrMany.php
@@ -42,7 +42,12 @@ trait AttachOneOrMany
         if (static::$constraints) {
             $this->query->where($this->morphType, $this->morphClass);
 
-            $this->query->where($this->foreignKey, '=', $this->getParentKey());
+            // Bind the parent key as a string to match the string `attachment_id` column. A numeric
+            // binding forces the database to coerce the column for every row, which prevents the
+            // column's index from being used.
+            $parentKey = $this->getParentKey();
+
+            $this->query->where($this->foreignKey, '=', is_null($parentKey) ? $parentKey : (string) $parentKey);
 
             $this->query->where('field', $this->getFieldName());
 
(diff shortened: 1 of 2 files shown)


austinderrick:
I ran this on a fresh MariaDB 10.11 database. It held 500,000 rows in system_files, built from Winter's own migration, where attachment_id is a varchar with its own index. On develop, the relation binds the parent key as an integer. So MariaDB compares a varchar to a number, and it has to coerce every row, which throws the index away. EXPLAIN confirms it: a full table scan, type ALL, key NULL, reading 496,451 rows.

The query averaged 115.5 ms over 100 runs. The string cast in this PR keeps the comparison string to string. Now the same query uses the index, type ref, and reads 5 rows. It averaged 0.37 ms, about 312 times faster. The real $user->avatar relation behaves the same way through Eloquent. The lazy load drops from 436 ms to 4 ms. The eager with('avatar') load drops from 88 ms to 4.5 ms. The gain grows with the table, since the old scan slows down with every new row while the index lookup stays flat.

image

[index.html](https://github.com/user-attachments/files/29311831/index.html)
[reproduce.sh](https://github.com/user-attachments/files/29311837/reproduce.sh)
[schema.sql](https://github.com/user-attachments/files/29311838/schema.sql)
[probe_attach.php](https://github.com/user-attachments/files/29311839/probe_attach.php)
[evidence_explain.txt](https://github.com/user-attachments/files/29311840/evidence_explain.txt)

PR #234, opened 2026-06-11
Cache reflection lookups in ExtendableTrait magic method handling

austinderrick:
## Problem

Every `__get`/`__set`/`__call` on a class using `ExtendableTrait` (i.e. every attribute access on every Database model) currently:

- builds a fresh `ReflectionClass` and recursively walks the parent chain in `extensionGetParentClass()` to find the first non-extendable ancestor,
- resolves the parent's `__get`/`__set`/`__call` `ReflectionMethod` **twice** per call — once in `extensionMethodExists()` and again in `extensionCallMethod()`,
- and, for classes with behaviors, constructs another `ReflectionClass` per extension per property read in `extendableIsAccessible()`.

None of this was memoized, although every one of these resolutions depends only on the class, never the instance. A microbenchmark of the `__get` path shows roughly a 10x overhead per attribute access (~30ns direct vs ~340ns through the reflection path) — and a typical request makes thousands of magic-method calls through models, components and widgets, plus GC pressure from the throwaway reflection objects.

## Fix

Three per-class static caches, following the same pattern as the existing `extendableCallStatic()` cache in the same file:

- `extensionGetParentClass()` caches the resolved parent reflection (or `false`) per class; the recursion's intermediate steps are split into `extensionResolveParentClass()` and are not cached.
- `extensionMethodExists()` / `extensionCallMethod()` share a cached `ReflectionMethod` handle per class and method. `extensionCallMethod()` keeps a fallback to an uncached `getMethod()` for non-public methods so its behaviour for direct, unguarded calls is unchanged (including the `ReflectionException` for unknown methods).
- `extendableIsAccessible()` caches the visibility check per class and property.

All cached values are class-structure facts that cannot change at runtime, so there is no invalidation concern; dynamic methods and properties continue to live in the per-instance `$extensionData` and are checked before any of these paths.

## Testing

New tests assert the parent reflection is reused across calls and instances (and that subclasses memoize independently), and that repeated magic access behaves consistently across instances. The full test suite passes — the Database model tests exercise the cached `__get`/`__set`/`__call` paths heavily.

## Summary by CodeRabbit

* **Refactor**
  * Improved application performance through internal optimization.
  * Enhanced reliability and consistency of magic method and property access behavior.

* **Tests**
  * Added comprehensive tests for magic method and property access patterns to ensure consistent behavior across repeated operations.

### DIFF
--- a/src/Extension/ExtendableTrait.php
+++ b/src/Extension/ExtendableTrait.php
@@ -50,6 +50,21 @@ trait ExtendableTrait
     protected static $extendableClassLoader = null;
 
     /**
+     * @var array Cache of parent class reflections (or `false` when there is none), keyed by class name.
+     */
+    protected static $extensionParentClassCache = [];
+
+    /**
+     * @var array Cache of magic method reflections (or `false` when unavailable), keyed by "Class::method".
+     */
+    protected static $extensionMethodReflectionCache = [];
+
+    /**
+     * @var array Cache of property accessibility checks, keyed by class name and property name.
+     */
+    protected static $extensionAccessiblePropertyCache = [];
+
+    /**
      * This method should be called as part of the constructor.
      */
     public function extendableConstruct()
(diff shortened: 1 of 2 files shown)


LukeTowers:
@austinderrick can you run performance testing on this to see how much it meaningfully impacts the performance of typical requests in terms of memory usage and requests per second that can be handled?

PR #238, opened 2026-08-09
Resolve Halcyon DB template mtimes without a query per check

LukeTowers:
## Problem

`Halcyon\Builder::getCached()` validates **every warm cache hit** by calling `isCacheBusted()`, which calls `datasource->lastModified()`. For database-backed templates that is a full `SELECT *` per template per request, purely to read `updated_at`:

```sql
select * from `cms_theme_templates` where `source` = ? and `deleted_at` is null and `path` = ? limit 1
```

So the Halcyon object cache removes **zero** database traffic for DB-backed templates — it only saves the parse. On a site taking heavy bot traffic this was the second-largest consumer of database time (717.8K calls, 2h 01m, 10.08ms avg — most of that cost being the `content` blob transfer).

## Fix

`getAvailablePaths()` already builds a `rememberForever` manifest of the paths held by the datasource in a **single** query, and `AutoDatasource` invalidates it on every insert/update/delete. Recording each live record's modification time in that manifest lets consumers resolve mtimes with no additional query.

Timestamps are truthy, so the existence/deletion contract of the map is unchanged and every existing consumer keeps working (`!$paths[$path]`, `array_filter`, `isset`). The paths cache key is versioned (`-v2-`) so stale boolean payloads are rebuilt rather than misread, and the documented `halcyon.datasource.db.beforeGetAvailablePaths` event may still return booleans.

Also narrows `lastModified()` to `select('updated_at')` instead of the whole row, for the paths that still query. The surrounding `try/catch` only existed to swallow the null-property error on a missing record, so it is replaced with an explicit check.

### Docblock

`getAvailablePaths()` previously documented (and inherited) a boolean-only contract. The return shape is now widened, so both `DatasourceInterface` and this implementation describe it explicitly as `array<string, int|bool>`. The `halcyon.datasource.db.beforeGetAvailablePaths` event may still supply plain booleans, so the timestamps are an optimization consumers must not rely on being present.

#### Keeping the manifest's truthiness invariant

Consumers decide whether a path can be handled by testing the manifest value for **truthiness** — `AutoDatasource::getDatasourceForPath()` treats a falsy value as deleted, `getValidPaths()` filters on it, `selectOne()` retries on it. Storing timestamps breaks that for exactly one value: a live record whose `updated_at` is the Unix epoch yields `0`.

Caught in review. Before the fix, such a record vanished:

| | before fix | after fix |
|---|---|---|
| manifest value | `0` | `true` |
| in `select()` listing | **no** | yes |
| `selectOne()` | **null** | returns content |
| `lastModified()` | `0` | `0` (live query) |

Note `lastModified()` still returned `0` while everything else treated the record as deleted — because it tests with `is_int()` while the others test truthiness. That asymmetry between a type test and a truthiness test is the actual defect.

Falling back to `true` for the epoch restores the invariant, so `false` is once again the only falsy value the map can hold, and such records simply resolve their mtime with a query as before. `DatasourceInterface` now documents the requirement so other implementations don't reintroduce it.

## Out of scope: nullable `updated_at`

`updated_at` is nullable, and `Carbon::parse(null)` resolves to *now* in `selectOne()` (which supplies the cached `mtime`), `lastModified()`, and the manifest alike. Those three never agree for such a record, so `isCacheBusted()` is always true and the cache is busted on every request — before and after this change. Verified:

```
selectOne mtime : 1786348933
manifest (t0)   : 1786348933
lastModified(t1): 1786348934
```

This is pre-existing and deliberately not addressed here. Records with a null `updated_at` cannot be produced through `insert()` or `update()`, which always set the column; they only arise from direct SQL or seeding. Fixing it properly means agreeing on one truthy sentinel across all three call sites (`0` would read as "deleted" in the manifest), which is a separate concern from this PR.

## Tests

Neither datasource had any coverage before this branch, so both are now covered properly rather than only the lines that moved — 68 tests across the two.

**`tests/Halcyon/DbDatasourceTest.php`** (36) — `selectOne`/`select` with column, extension and fileMatch filters; insert/update/delete/forceDelete including renames, extension changes and reviving soft-deleted records; `lastModified`; the paths manifest; cache keys; and the DB-specific behaviours: soft deletion, source scoping, and the `beforeInsert` / `beforeUpdate` / `extendQuery` / `beforeGetAvailablePaths` events.

**`tests/Halcyon/FileDatasourceTest.php`** (32) — the same interface surface, plus nested directory creation, path traversal handling, `maxDepth` recursion and live mtimes.

### The one that matters for reviewers

`FileDatasource::getAvailablePaths()` deliberately **still returns `true`**, not a timestamp, and `testGetAvailablePathsReturnsTrueForEveryPath` pins that. Reporting timestamps there would be actively harmful — unlike a database round trip, a filesystem mtime is a cheap local stat, and freezing it into the forever-cached manifest would mean template edits on disk (a deploy, say) go unnoticed until the manifest is rebuilt. `AutoDatasource` depends on the `true` value to fall through to a live `filemtime()`.

Verified by patching `FileDatasource` to return `filemtime()` — the test fails immediately.

### Regression guards

Run against the pre-change source (`fd673f4f`), three fail and the other 33 pass, so the rest genuinely characterize existing behaviour:

```
1) testGetAvailablePathsReturnsTimestampsForLiveRecords
   Failed asserting that true is identical to 1559390400.
2) testLastModifiedDoesNotSelectTheContentColumn
   Failed asserting that 'select * from "halcyon_tester_templates" where "source" = ? and "deleted_at" is null and "path" = ? limit 1' does not contain "*".
3) testGetPathsCacheKeyIsVersionedAndScoped
   Failed asserting that two strings are identical.
```

Two pre-existing behaviours are asserted as they actually are rather than as they might be assumed: `FileDatasource::selectOne()` swallows the traversal exception along with every other read error, so escaping the base path returns `null` rather than throwing; and `DbDatasource::delete()` throws for an already soft-deleted record, because its query excludes them.

Full suite: 840 tests. The 1 error / 4 failures that remain are pre-existing (`ConfigWriterTest`, `ArrayFileTest` — PHP config-file formatting) and are present on `develop` without this change.

## Companion PR

Requires wintercms/winter#1508, which consumes the manifest in `Cms\Classes\AutoDatasource::lastModified()`. Land together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing records by returning no modification date instead of an error.
  * Available paths now show accurate update timestamps for active records.
  * Deleted paths continue to be identified correctly.
  * Database lookups are properly scoped and avoid loading unnecessary content.

* **Tests**
  * Added coverage for timestamps, deleted records, source scoping, event filtering, and missing records.

### DIFF
--- a/src/Halcyon/Datasource/DatasourceInterface.php
+++ b/src/Halcyon/Datasource/DatasourceInterface.php
@@ -118,19 +118,23 @@ interface DatasourceInterface
     /**
      * Get all available paths within this datasource.
      *
-     * This method returns an array, with all available paths as the key, and a boolean that represents whether the path
-     * can be handled or modified.
+     * This method returns an array, with all available paths as the key, and a value that represents whether the path
+     * can be handled or modified. A falsy value means the path cannot be handled; any truthy value means it can.
+     *
+     * Datasources that are able to determine a path's modification time cheaply may return that timestamp as the
+     * truthy value, allowing consumers to resolve modification times without a further lookup.
      *
      * Example:
      *
      * ```php
      * [
      *     'path/to/file.md' => true, // (this path is available, and can be handled)
-     *     'path/to/file2.md' => false // (this path is available, but cannot be handled)
+     *     'path/to/file2.md' => 1559390400, // (as above, and was last modified at this timestamp)
+     *     'path/to/file3.md' => false // (this path is available, but cannot be handled)
      * ]
      * ```
      *
-     * @return array An array of available paths alongside whether they can be handled.
+     * @return array<string, int|bool> An array of available paths alongside whether they can be handled.
      */
     public function getAvailablePaths(): array;
 }
--- a/src/Halcyon/Datasource/DbDatasource.php
+++ b/src/Halcyon/Datasource/DbDatasource.php
@@ -341,25 +341,37 @@ class DbDatasource extends Datasource
      */
     public function lastModified(string $dirName, string $fileName, string $extension): ?int
     {
-        try {
-            return Carbon::parse($this->getQuery()
-                    ->where('path', $this->makeFilePath($dirName, $fileName, $extension))
-                    ->first()->updated_at)->timestamp;
-        } catch (Exception $ex) {
-            return null;
-        }
+        $record = $this->getQuery()
+            ->select('updated_at')
+            ->where('path', $this->makeFilePath($dirName, $fileName, $extension))
+            ->first();
+
+        return $record ? Carbon::parse($record->updated_at)->timestamp : null;
     }
 
     /**
      * @inheritDoc
+     *
+     * The key is versioned because the payload shape has changed; see getAvailablePaths().
      */
     public function getPathsCacheKey(): string
     {
-        return 'halcyon-datastore-db-' . $this->table . '-' . $this->source;
+        return 'halcyon-datastore-db-v2-' . $this->table . '-' . $this->source;
     }
 
     /**
-     * @inheritDoc
+     * Get all available paths within this datasource.
+     *
+     * Live records are mapped to their last modification timestamp rather than `true` so that
+     * consumers can resolve mtimes without a further query. Timestamps are truthy, so the
+     * existence / deletion contract of this map is unchanged.
+     *
+     * Note that the `halcyon.datasource.db.beforeGetAvailablePaths` event below may still
+     * return plain booleans, so consumers must treat the timestamps as an optimization rather
+     * than something they can rely on being present.
+     *
+     * @return array<string, int|bool> Paths that cannot be handled are `false`; live paths are
+     *                                 a timestamp, or `true` when supplied by the event below.
      **/
     public function getAvailablePaths(): array
     {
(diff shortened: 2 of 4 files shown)


LukeTowers:
Thanks both. Addressed one of the two, and deliberately not the other.

## Docblock / `@inheritDoc` — fixed

Fair catch. `getAvailablePaths()` inherited a docblock that explicitly documents a boolean map, complete with an example, so the widened return shape was genuinely misleading for phpstan and IDEs. Both `DatasourceInterface` and the `DbDatasource` implementation now describe it as `array<string, int|bool>`.

Worth noting for anyone consuming this: the `halcyon.datasource.db.beforeGetAvailablePaths` event may still return plain booleans, and `FileDatasource` returns `true`, so the timestamps are an optimization that consumers must not rely on being present. That is now stated in both docblocks.

## Nullable `updated_at` — real, but pre-existing and out of scope

The underlying defect is real, though it is worse than described. It is not just that `lastModified()` and the manifest can drift; `selectOne()` also resolves `Carbon::parse(null)` to *now*, and that is what supplies the cached `mtime` that `isCacheBusted()` compares against. So all three disagree:

```
selectOne mtime : 1786348933
manifest (t0)   : 1786348933
lastModified(t1): 1786348934
```

The consequence is that a record with a null `updated_at` has its cache busted on **every** request and never converges — both before and after this PR. So this is not a regression introduced here.

Not fixing it in this PR, for two reasons:

1. Such records cannot be produced through the datasource API. Both `insert()` and `update()` always set `updated_at`; nulls only arise from direct SQL or seeding.
2. A correct fix has to apply one shared sentinel across all three call sites, not just the two flagged. It also has to be **truthy** — `0` or `null` would read as "cannot be handled" in the manifest and make the path look deleted, which is precisely the contract this PR is careful to preserve.

That is a self-contained change with its own reasoning and test surface, so it belongs in its own PR rather than being bolted onto this one.

@coderabbitai one correction: the manifest does not "return different mtimes across reads without a database update". It is built once and held in `Cache::rememberForever`, so it is stable for the lifetime of that entry. It is only rebuilt per-request when `app.debug` is enabled, because `AutoDatasource::fetchPathCache()` intentionally bypasses the cache in debug mode — dev-only, and pre-existing.

I have also removed the "incidental fix" claim from the PR description, which overstated what freezing the manifest value achieves.

PR #242, opened 2026-08-14
Reset the force-deleting flag when a force delete throws

austinderrick:
If `delete()` throws inside `forceDelete()`, the flag never resets and the datasource is stuck in force-delete mode. Every ordinary delete after that runs as a hard delete.

The fix wraps the call in try/finally so the flag always resets and the exception still reaches the caller. Two tests cover the throwing path and the happy path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved force deletion cleanup so its internal state is reset even when deletion fails.
  - Deletion errors continue to be reported correctly, while successful deletions retain their existing results.

- **Tests**
  - Added coverage for successful and failed force-deletion scenarios.

### DIFF
--- a/src/Halcyon/Datasource/Datasource.php
+++ b/src/Halcyon/Datasource/Datasource.php
@@ -59,11 +59,13 @@ abstract class Datasource implements DatasourceInterface
     {
         $this->forceDeleting = true;
 
-        $success = $this->delete($dirName, $fileName, $extension);
-
-        $this->forceDeleting = false;
-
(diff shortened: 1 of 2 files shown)


LukeTowers:
@austinderrick if you retarget develop I'll merge this in before I release 1.2.14 later this evening or this weekend.

austinderrick:
@LukeTowers : Updated!

LukeTowers (approved):

